PR #11, opened 2016-07-28
Adding new design towards telegram

sudheesh001:
This is a barebone we can continue working with

screen shot 2016-07-28 at 11 39 54 pm


mariobehling:
This is an older commit, that we did not follow up. As it gets more busy here again, let's try it out.

Issue #16, opened 2016-08-12
Susi interface in a Linux text Browser?


smokingwheels:
What about having Susi running in a Linux text Browser? [Here](https://twitter.com/smokingwheels/status/764035564832116736)

I am not sure really what the problem is but I cant seem to get it to work.

SpEcHiDe:
I think the problem is JavaScript being disabled by default in most text based browsers. 
Please read here: http://elinks.or.cz/documentation/manual.html#ecmascript or we may need to think some other method to show a "Switch to modern browser warning" when opened on text based browsers! 

I think, this can be done by adding a <noscript> tag!

PR #34, opened 2016-12-10
add fallback instructions for a text based browser issue #16

SpEcHiDe:


mariobehling:
@SpEcHiDe Thanks for getting on more. Please help us and follow best practices when committing. This will make it much easier for everyone to review.

1. Open an issue for anything your encounter.
2. Make a PR and add information with a link to the issue in the description.
3. Add a screenshot and a link to a setup we can test, if possible.

Thank you.

SpEcHiDe:
I assumed, since there is already an issue https://github.com/fossasia/asksusi.com/issues/16 here, I need not create a new issue, and make it redundant. 
I have not setup the development environment, so I can't give a link to the setup. 
But, I assume, by reading the documentation available on [MDN](https://developer.mozilla.org/), that merely adding a <noscript> tag inside the body, would cause text based browsers to display that text, and the browsers with javaScript enabled would not see anything inside the <noscript> tag. The above is assumed from my understanding of the documentation available at [MDN](https://developer.mozilla.org/)

mariobehling:
@SpEcHiDe Thanks. Yes, you are right the issue #16 is good enough. Just missed it in the description then. Cheers.

PR #80, opened 2017-05-18
Fixes #79 - Proccessing susi response for hyperlinking urls

### LINKED ISSUE #79: Processing response for hyperlinking urls
If the response from susi contains urls, they could be replaced by a phrase like "Click Here" hyperlinked to that url.

uday96:
Proccesses the responses from susi and replaces any urls with "Click Here" hyperlinking that url

### DIFF
--- a/src/components/MessageListItem.react.js
+++ b/src/components/MessageListItem.react.js
@@ -1,17 +1,23 @@
 import React from 'react';
 import { PropTypes } from 'prop-types';
+import Linkify from 'react-linkify';
+
+export const parseAndReplace = (text) => {return <Linkify properties={{target:"_blank"}}>{text}</Linkify>;}
 
 class MessageListItem extends React.Component {
 
   render() {
     let {message} = this.props;
+    let stringWithLinks = this.props.message.text;  
+    let replacedText = parseAndReplace(stringWithLinks);
+    
     return (
       <li className="message-list-item">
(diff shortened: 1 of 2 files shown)


rishiraj824:
@uday96 Can you provide a demo link?

src/components/MessageListItem.react.js
```diff
@@ -1,17 +1,35 @@
 import React from 'react';
 import { PropTypes } from 'prop-types';
+const processString = require('react-process-string');
```
rishiraj824: Also try using ```import processString from 'react-process-string';``` to import files following the ES6 syntax.

uday96:
@rishiraj824 @isuruAb will make a pr using a different module, as pointed out react-process-string module is causing the build to fail due to incompatible syntax :+1:

rishiraj824:
@uday96 Please squash the commits. And provide the demo link.

uday96:
@isuruAb  @rishiraj824 Im having connectivity issues deploying it for a demo link.
http://webchatdemo.surge.sh/ is the link for the previous commit. Here the url opens in the same tab. I changed that by adding target="_blank" in the recent commit. WIll post a demo link for the recent one soon.
Also, All kinds of url links and also mail ids are hyperlinked but I couldnt replace it with a phrase like "Click Here" . WIll work on that too.

Issue #84, opened 2017-05-19
SUSI does not show the answer until we ask another question 


isuruAb:
I have tried with several questions.it does not show the answers until we ask another question.
<img width="781" alt="screen shot 2017-05-19 at 9 50 31 pm" src="https://cloud.githubusercontent.com/assets/7692626/26257190/4d258226-3cdd-11e7-8b80-8d1b0d6540ff.png">
 
after we ask another question it suddenly show the answer for previous question like this.
<img width="819" alt="screen shot 2017-05-19 at 9 52 42 pm" src="https://cloud.githubusercontent.com/assets/7692626/26257265/8f1da3ca-3cdd-11e7-94e0-0377e808a966.png">

rishiraj824:
Will work on this :+1:

PR #87, opened 2017-05-20
Fixes #71 - Map support for webchat app

### LINKED ISSUE #71: Implement support for maps in chat client
Map support is required for some answer types.

uday96:
http://susimaps.surge.sh/ is the demo link.
Try querying "where is london" to check the response with mapsupport.
The marker icon is not rendering due to error in leaflet source code. Working on a workaround or a custom marker icon.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,11 +9,13 @@
     "flux": "^3.1.2",
     "jquery": "^3.2.1",
     "keymirror": "^0.1.1",
+    "leaflet": "^1.0.3",
     "material-ui": "^0.18.0",
     "prop-types": "^15.5.10",
     "react": "^15.5.4",
     "react-bootstrap": "^0.31.0",
     "react-dom": "^15.5.4",
+    "react-leaflet": "^1.1.6",
     "react-linkify": "^0.2.1",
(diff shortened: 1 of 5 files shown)


rishiraj824:
Every time, after sending the second message the map response is generated. This issue(https://github.com/fossasia/chat.susi.ai/issues/84) needs to be fixed first in order to fix the PR.

It's because of the latency which is added `let NETWORK_LATENCY = 1300;` in this [file](https://github.com/fossasia/chat.susi.ai/blob/master/src/ChatExampleDataServer.js). So if the response time of loading a map increases 1300ms the answer would load but it won't show up until the callback function is called since the latency is allowed for only 1300ms. The map loads only after an avg. of 2400ms -2700ms. So we need to figure out a way that the responses show up as soon as the server sends the response.
```
export function getMessages(callback) {
  setTimeout(() => {
    callback(messages);
  }, NETWORK_LATENCY);
};

```

rishiraj824:
Made PR #88, it will fix the above issue. @uday96

uday96:
@daminisatya @rishiraj824 updated my branch and the demo link. please review.

Issue #85, opened 2017-05-19
Can we use Semantic UI for this project


mariobehling:
Semantic UI already has a lot of components we could use. Please check if we can use it.

https://semantic-ui.com/

rishiraj824:
@mariobehling I think we can, but the components will look similar to the current one on http://chat.susi.ai/chat
We can definitely proceed with this if there's no problem with that. Or the second option can be the Android UI mentioned at https://github.com/fossasia/chat.susi.ai/issues/69

isuruAb:
Semantic UI is good but it's components are just like normal UI elements. here they compare bootstrap , material ui and semantic web https://stackshare.io/stackups/bootstrap-vs-material-ui-vs-semantic-ui
here they have compared material UI and semantic UI:  https://www.slant.co/versus/520/521/~semantic-ui_vs_materialize

BrainBuzzer:
Semantic UI comes with great theming support. You can customize almost every single bit of a semantic UI theme. Leave that to me.

isuruAb:
this can be closed because now we are using material-UI as our UI library

PR #89, opened 2017-05-22
Fixes #77 - Table support for chat client

### LINKED ISSUE #77: Implement Table view for table action types
http://api.susi.ai/susi/chat.json?q=stock+quotes currently gives a table in the following format. Implement a table for the same in the response.
```
{
  "query": "stock quotes",
  "answers": [{
    "data": [
      {"currency": "EUR/USD", "rate": "1.09302", "timestamp": "1494621900069"},
      {"currency": "USD/JPY", "rate": "113.326", "timestamp": "1494621900186"},
      {"currency": "GBP/USD", "rate": "1.28835", "timestamp": "1494621900129"},
      {"currency": "EUR/GBP", "rate": "0.84831", "timestamp": "1494621900103"},
      {"currency": "USD/CHF", "rate": "1.00133", "timestamp": "1494621899461"}
    ],
    "metadata": {"count": 59},
    "actions": [{
      "type": "table",
      "columns": {
        "currency": "Valuta",
        "rate": "Quota"
      }
    }]
  }]
}
```

uday96:
demo link : http://susitables.surge.sh/
sample query : universities in germany

### DIFF
--- a/src/components/MessageListItem.react.js
+++ b/src/components/MessageListItem.react.js
@@ -2,6 +2,15 @@ import React from 'react';
 import { PropTypes } from 'prop-types';
 import { Map, Marker, Popup, TileLayer } from 'react-leaflet';
 import Linkify from 'react-linkify';
+import {
+  Table,
+  TableBody,
+  TableHeader,
+  TableHeaderColumn,
+  TableRow,
+  TableRowColumn,
+} from 'material-ui/Table';
+import MuiThemeProvider from "material-ui/styles/MuiThemeProvider";
 
 export const parseAndReplace = (text) => {return <Linkify properties={{target:"_blank"}}>{text}</Linkify>;}
 
@@ -15,7 +24,7 @@ function drawMap(lat,lng){
       />
       <ExtendedMarker position={position}>
         <Popup>
-          <span>Hello! <br/> I'm here.</span>
+          <span>Hello! <br/> I am here.</span>
         </Popup>
       </ExtendedMarker>
     </Map>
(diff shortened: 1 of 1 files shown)


src/components/MessageListItem.react.js
```diff
@@ -45,6 +54,41 @@ class MessageListItem extends React.Component {
         let tableData = data.answers[0].data;
```
daminisatya: fix these lines

uday96: @daminisatya I guess the highlighting is because its in ES6 and the editor needs a babel plugin to support it. The code works fine. Did you mean to fix that only?

src/components/MessageListItem.react.js
```diff
@@ -2,6 +2,15 @@ import React from 'react';
...
 import { Map, Marker, Popup, TileLayer } from 'react-leaflet';
 import Linkify from 'react-linkify';
+import {
+  Table,
+  TableBody,
+  TableHeader,
+  TableHeaderColumn,
+  TableRow,
+  TableRowColumn,
+} from 'material-ui/Table';
```
rishiraj824: Are we using material-ui #85 ?

uday96: Sorry I missed that discussion. We dint use any other material UI modules? So do we let it be till we decide on something?

uday96:
@rishiraj824 Oh okay. I'll fix it :)

AnshulMalik:
@rishiraj824 @uday96 let statements has nothing to do with the red lines.
(Here)[https://stackoverflow.com/questions/34570348/what-is-the-meaning-of-these-red-bars-in-git-file-difference] is a nice explanation.

src/components/MessageListItem.react.js
rishiraj824: @uday96 We can work through the UI in the later issues. Works for now :+1:

Issue #70, opened 2017-05-13
Setup auto-deployment in gh-pages branch


mariobehling:
Please set up autodeployment to chat.susi.ai.

rishiraj824:
@mariobehling can be closed now.

rishiraj824:
Closing this by #93

PR #94, opened 2017-05-24
Updated deploy key

rishiraj824:
The build was failing, now updated the key. Hope it works
@mariobehling


AnshulMalik:
Fixes #70

PR #97, opened 2017-05-26
Fixes #81 create some test case

### LINKED ISSUE #81: Create Test Cases for the App
Currently, there are no test cases for all the issues. As the App is being made one needs to keep track of all the changes and run proper tests to ensure everything is working.

isuruAb:
created few test cases for components and actions.
resolve #81


AnshulMalik:
Great work @isuruAb 

How about we add some more tests to check the content of components, like classes and elements, by passing custom state and props.

src/__tests__/components/MessageComposer.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import MessageComposer from '../../components/MessageComposer.react';
+import { shallow } from 'enzyme';
+
+ it('render MessageComposer without crashing',()=>{
+   shallow(<MessageComposer />);
+ })
```
rishiraj824: Should add a new line after the end of the file.

src/__tests__/components/ThreadSection.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import ThreadSection from '../../components/ThreadSection.react';
+import { shallow } from 'enzyme';
+
+ it('render ThreadListItem without crashing',()=>{
+   shallow(<ThreadSection />);
+ })
```
rishiraj824: Should add a new line after the file.

src/__tests__/utils/ChatMessageUtils.js
```diff
@@ -0,0 +1,59 @@
...
+
+it('check getSUSIMessageData util', () => {
+    var timestamp = Date.now();
+    var message = {
+        id: 'm_' + timestamp,
+        threadID: "t_1",
+        authorName: 'SUSI', // hard coded for the example
+        text: 'Hello',
+        response: {},
+        date: new Date(timestamp),
```
AnshulMalik: timestamp is already Date object ?

src/__tests__/utils/ChatMessageUtils.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+    expect(ChatMessageUtils.getSUSIMessageData(message, "t_1").id).toBe(message.id);
+    expect(ChatMessageUtils.getSUSIMessageData(message, "t_1").threadID).toBe(message.threadID);
+    expect(ChatMessageUtils.getSUSIMessageData(message, "t_1").authorName).toBe(message.authorName);
+    expect(ChatMessageUtils.getSUSIMessageData(message, "t_1").text).toBe(message.text);
+    expect(ChatMessageUtils.getSUSIMessageData(message, "t_1").response).toBe(message.response);
+});
+
+it('check getCreatedMessageData util', () => {
+    var timestamp = Date.now();
```
AnshulMalik: I'd suggest `const` instead of `var`

src/__tests__/utils/ChatMessageUtils.js
```diff
@@ -0,0 +1,59 @@
...
+    var messageText = "Hello";
+    var currentThreadID = "t_1";
+    var message = {
+        id: 'm_' + timestamp,
+        threadID: currentThreadID,
```
AnshulMalik: I think this would just increase redundancy, I mean duplicate code for message, what do you think ?  @isuruAb 
If we change structure of message in utils, we need to change this as well.

src/__tests__/utils/ChatMessageUtils.js
```diff
@@ -0,0 +1,59 @@
...
+        authorName: 'You',
+        date: new Date(timestamp),
+        text: messageText,
+        isRead: true
+    }
+    expect(ChatMessageUtils.getCreatedMessageData(messageText, currentThreadID).id).toBe(message.id);
```
AnshulMalik: We can do better here, 
```
const chatMessage = ChatMessageUtils.getCreatedMessageData(messageText, currentThreadID);
expect(chatMessage.threadID).toBe(message.threadID);
```

Looks cleaner this way. Similarly for other tests.

isuruAb: this one is really good. i will change this. thanks for pointing out this

src/__tests__/utils/ChatMessageUtils.js
isuruAb: @AnshulMalik  but the case is different functions contains different objects.thats why I added separate objects

AnshulMalik: ok

PR #101, opened 2017-05-28
Fixes #100 : Add support for eslint

### LINKED ISSUE #100: setup ESLint for Application
With ESLint we can verify code quality and  minimise errors that can be occured because of poor quality codes.
http://eslint.org/

AnshulMalik:
I have added some basic rules so that we can start linting, we can iterate from here.

### DIFF
--- a/src/components/MessageListItem.react.js
+++ b/src/components/MessageListItem.react.js
@@ -10,10 +10,14 @@ import {
   TableRow,
   TableRowColumn,
 } from 'material-ui/Table';
-import MuiThemeProvider from "material-ui/styles/MuiThemeProvider";
+import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
 import { divIcon } from 'leaflet';
 
-export const parseAndReplace = (text) => {return <Linkify properties={{target:"_blank"}}>{text}</Linkify>;}
+export function parseAndReplace(text) {
+  return <Linkify properties={{target: '_blank'}}>
+    {text}
+  </Linkify>;
+}
 
 function drawMap(lat,lng){
   let position = [lat, lng];
@@ -38,10 +42,10 @@ function drawMap(lat,lng){
 }
 
 class ExtendedMarker extends Marker {
-  
+
   componentDidMount() {
     super.componentDidMount();
-    
+
     this.leafletElement.openPopup();
   }
 }
@@ -50,7 +54,7 @@ class MessageListItem extends React.Component {
 
   render() {
     let {message} = this.props;
-    let stringWithLinks = this.props.message.text;  
+    let stringWithLinks = this.props.message.text;
     let replacedText = parseAndReplace(stringWithLinks);
 
     if(this.props.message.hasOwnProperty('response')){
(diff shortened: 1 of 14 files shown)


src/components/MessageListItem.react.js
```diff
@@ -61,24 +65,33 @@ class MessageListItem extends React.Component {
...
-                return(<TableRowColumn><Linkify properties={{target:"_blank"}}>{eachrow[key]}</Linkify></TableRowColumn>);
+                return <TableRowColumn key={key}>
+                  <Linkify properties={{ target: '_blank' }}>
+                    {eachrow[key]}
+                  </Linkify>
+                </TableRowColumn>;
               });
               return(
-                <TableRow>{rowcols}</TableRow>
+                <TableRow key={i}>{rowcols}</TableRow>
```
isuruAb: what is this change.is it related to eslint issue ?

AnshulMalik:
Yes, this is recommended for react components​ to use key for each of
component that an iterator returns.
So it's in enabled in eslint as well.

src/stores/MessageStore.js
```diff
@@ -9,17 +9,6 @@ let CHANGE_EVENT = 'change';
 
 let _messages = {};
 
-function _addMessages(rawMessages) {
```
isuruAb: could you please explain why you delete this snippet ?

AnshulMalik: This wasn't used anywhere, so thought it's not required.

isuruAb: @AnshulMalik  good but it is not related to #100 issue.if we can do only related changes in commits.that will be really easy to others.

AnshulMalik: Yes @isuruAb, that's right, we shouldn't mix those two, but this was an eslint issue as well, it's an coincidence that it belongs to two kinds of issues.

src/components/MessageListItem.react.js
AnshulMalik: Again, I have done only the changes that the eslint config required, so our existing codebase was having issues with linting, and I have resolved those just to pass the tests.

Otherwise, I'd have to fix it first, create a PR and then add another rule to eslint. I thought of adding this change to initial eslint PR.

src/stores/MessageStore.js
rishiraj824: @AnshulMalik This actually converts the user's rawMessage to a message model which is defined in actions.js It is required.

isuruAb: @AnshulMalik  for now it is okay. btw awesome commit 👍

AnshulMalik: @rishiraj824 I am still not able to find the usage of `_addMessages` in actions.js, can you point me to the line please?

mariobehling:
Please resolve merge conflicts.

uday96:
@AnshulMalik Resolve merge conflicts :)

src/stores/MessageStore.js
rishiraj824: @AnshulMalik Every Message is being mapped to a  _messages array and this is also being used to update the latest timestamp if you look at the function it is calling ie. ChatMessageUtils.convertRawMessage()
```
export function convertRawMessage(rawMessage, currentThreadID) {
  return {
    ...rawMessage,
    date: new Date(rawMessage.timestamp),
    isRead: rawMessage.threadID === currentThreadID
  };
};
```
Also when i tried deleting it the chat messages weren't being updated on the chat window, which is not surprising  at all because of course it needs to be added to the array of messages right.

AnshulMalik: Thanks for the review, I rechecked, this method was not used anywhere in the code and doesn't affect any other component. Everything works fine when this function is removed. 
Please check by pulling my PR.

Thanks

PR #109, opened 2017-05-30
Fixes #78 - websearch support for chat client

### LINKED ISSUE #78: Implement web result for unknown responses.
For the responses like **I don't know how to answer this. Here is a web search result:** show the web result or a generate a hyperlink on the query.
issue1

uday96:
**Demo Link:** http://susiwebsearch.surge.sh/
**Sample Queries:** 'Germany', 'Marvel', 'artificial Intelligence' etc which give action type websearch

For few queries like 'find me restaurants', even duckduckgo doesnt give any response. How to handle such cases?

Also the css can be changed. And the number of results too.
Right now, Im displaying all results without 'Name' attribute i.e popular results and one from each category (when 'Name' attribute is present).

ai

I refactored MessageListItem.js as I added actions attribute to message. I know its not related to this issue, will take care from next time :)

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -1,0 +1,23 @@
+**Actual Behaviour**
+
+Please state here what is currently happening.
+
+**Expected Behaviour**
+
+State here what the feature should enable the user to do.
+
+**Steps to reproduce it**
+
+Add steps to reproduce bugs or add information on the place where the feature should be implemented. Add links to a sample deployment or code.
+
+**LogCat for the issue**
+
+Provide logs for the crash here
+
+**Screenshots of the issue**
+
+Where-ever possible attach a screenshot of the issue.
+
+**Would you like to work on the issue?**
+
+Please let us know if you can work on it or the issue should be assigned to someone else.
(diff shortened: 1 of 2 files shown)


src/actions.js
```diff
@@ -60,21 +62,50 @@ export function createSUSIMessage(createdMessage, currentThreadID) {
...
+          let actionIndex = actions.indexOf('websearch');
+          let query = response.answers[0].actions[actionIndex].query;
+          $.ajax({
+            url: 'http://api.duckduckgo.com/?format=json&q='+query,
+            dataType: 'jsonp',
+            crossDomain: true,
+            timeout: 3000,
+            async: false,
+            success: function (data) {
+                receivedMessage.websearchresults = data.RelatedTopics;
```
rishiraj824: Can we have the Abstract to be the first card with the image. And then the RelatedTopics? @uday96

uday96: Sure, that would be more informative. Will make changes :+1:

uday96:
@rishiraj824 Added the abstract if present as the first tile. But the css for the card is not complete. I hardcoded the height as 150px. Must make it auto size and render the contents accordingly.

uday96:
Messed up while squashing commits :sob: 
PR at https://github.com/fossasia/chat.susi.ai/pull/111

PR #111, opened 2017-05-30
Fixes #78 (partly) - Websearch support

### LINKED ISSUE #78: Implement web result for unknown responses.

uday96:
Demo Link: http://susiwebsearch.surge.sh/
Sample Queries: 'Germany', 'Marvel', 'artificial Intelligence' etc which give action type websearch

For few queries like 'find me restaurants', even duckduckgo doesnt give any response. How to handle such cases?

Also the css can be changed. And the number of results too.
Right now, Im displaying all results without 'Name' attribute i.e popular results and one from each category (when 'Name' attribute is present).

ai

I refactored MessageListItem.js as I added actions attribute to message. I know its not related to this issue, will take care from next time :)

Also added the abstract if present as the first tile - but needs css support

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -47,6 +47,8 @@ export function createSUSIMessage(createdMessage, currentThreadID) {
     authorName: 'SUSI', // hard coded for the example
     text: '',
     response: {},
+    actions:[],
+    websearchresults:[],
     date: new Date(timestamp),
     isRead: true,
   };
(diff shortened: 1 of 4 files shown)


AnshulMalik:
How about we show a more relevant title instead of Category, as user might not be so much interested in category, it should be below the Title, like
Title
Category
Description

uday96:
@AnshulMalik What to put under title? The API response doesnt have anything which we could use to put as title. Incase you find something, it can be included in another PR which fixes CSS for this :smile:

PR #114, opened 2017-05-31
Fixes #113: Changed few files according to best practises

### LINKED ISSUE #113: Refactor code base according to best practises
**Actual Behaviour**

Currently, there are very few components of the app not following best practices.

**Expected Behaviour**

We should always follow best practices. l found this link useful for following best practices :https://github.com/airbnb/javascript
In future, we should follow this only.

Yes, sending PR.

madhavrathi:
Fixes issue #113 

Changes: changed few files in accordance with best practices.

### DIFF
--- a/src/components/MessageComposer.react.js
+++ b/src/components/MessageComposer.react.js
@@ -1,11 +1,11 @@
 
 import * as Actions from '../actions';
-import React from 'react';
+import React,{Component} from 'react';
 import { PropTypes } from 'prop-types';
 
 let ENTER_KEY_CODE = 13;
 
-class MessageComposer extends React.Component {
+class MessageComposer extends Component {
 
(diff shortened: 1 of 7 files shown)


src/components/MessageComposer.react.js
```diff
@@ -1,11 +1,11 @@
 
 import * as Actions from '../actions';
-import React from 'react';
+import React,{Component} from 'react';
 import { PropTypes } from 'prop-types';
 
 let ENTER_KEY_CODE = 13;
 
-class MessageComposer extends React.Component {
+class MessageComposer extends Component {
```
AnshulMalik: Thanks for the work, but you should consider other best practices as this change, like changing components to lightweight stateless components where we don't require state :)

PR #131, opened 2017-06-02
Fixes #118 dark theme and ability to switch

### LINKED ISSUE #118: Two themes for chat application
**Actual Behaviour**

Now the web chat application has only one theme "light theme". 

**Expected Behaviour**

If we can implement a dark theme and way to switch between those two themes that would be really helpful for users who like to dark theme.

I would like to work on this issue

isuruAb:
Fixes issue #118 

**Changes:**
Added a button to switch between two themes.
made a store for settings for future use.

Demo Link:  http://susi-patch-118.surge.sh/

Screenshots for the change: 
screen shot 2017-06-03 at 4 44 11 am

### DIFF
--- a/src/components/ChatApp.css
+++ b/src/components/ChatApp.css
@@ -1,4 +1,15 @@
-
+.dark-body{
+  background-repeat: no-repeat;
+background: #183850;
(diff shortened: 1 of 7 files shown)


rishiraj824:
Some checks are failing @isuruAb

src/components/ChatApp.css
```diff
@@ -137,7 +170,10 @@ header{
     top: 16px;
     border-color: transparent transparent #d3e2f4 #d3e2f4;
     background: #d3e2f4;
-
+}
+.dark .message-container.SUSI::after{
```
rishiraj824: instead of using the CSS variables we can have sass variables in a sass file. which can then be used to make different themes for SUSI what do you think? same for the rest of the CSS

madhavrathi:
@isuruAb please squash your commits. Great work.

PR #148, opened 2017-06-04
Fixes #143 Add Material UI App bar

### LINKED ISSUE #143: Add Material UI App Bar
**Actual Behaviour**

Now the top bar that shows "SUSI" is made with raw HTML and CSS

**Expected Behaviour**

we can use Material UI App Bar instead of existing bar.
material UI app bar : http://www.material-ui.com/#/components/app-bar
current App-bar:
screen shot 2017-06-04 at 12 58 54 pm

I would like to work on this

isuruAb:
Fixes issue #143 
Changes:  add material UI App bar and option for switch between themes.

Demo Link: http://susi-patch-143.surge.sh/
Screenshots for the change: 
screen shot 2017-06-05 at 12 30 15 am

### DIFF
--- a/src/components/ChatApp.css
+++ b/src/components/ChatApp.css
@@ -79,6 +79,9 @@ nav h1 {
     font-weight: 900;
     color: #fff;
 }
+.app-bar h1{
+  text-align: center;
+}
 .dark nav h1{
     color: #fff;
 }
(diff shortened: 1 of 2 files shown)


src/components/ChatApp.css
```diff
@@ -79,6 +79,9 @@ nav h1 {
     font-weight: 900;
     color: #fff;
 }
+.app-bar h1{
+  text-align: center;
```
rishiraj824: Lets have the Lato font and the font weight as earlier @isuruAb

sudheesh001:
Looks good to me, If you want to go ahead with the Lato font please do that.

PR #159, opened 2017-06-06
Fixes #120: Route for accessing dream

### LINKED ISSUE #120: Make a route to directly access a dream
Can we have a route that loads a dream directly when the URL is called. for example when we call `chat.susi.ai/dream/wiki ` the web client should load the dream wiki.

rishiraj824:
Fixes issue #120 

Changes: 
- Added routes to test any dream from query

Demo Link: 
http://susi.surge.sh/?dream=susi

Param dream can be changed to access any dream

@uday96 @isuruAb @madhavrathi Please review

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "bootstrap": "^3.3.7",
     "classnames": "^2.2.5",
     "flux": "^3.1.2",
+    "history": "^4.6.1",
     "jquery": "^3.2.1",
     "keymirror": "^0.1.1",
     "leaflet": "^1.0.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -7,6 +7,7 @@
     "bootstrap": "^3.3.7",
     "classnames": "^2.2.5",
     "flux": "^3.1.2",
+    "history": "^4.6.1",
```
isuruAb: what this package does ?

rishiraj824: This is required since I am not using react-router, to instantiate the history manually. This is also suggested in the package that I am using.

Issue #151, opened 2017-06-05
Loop through action types and display result for all actions


chiragw15:
**Actual Behaviour**

Query "multi3". The sever response contains 3 action types but only result of first is displayed.
Also for map type only first action type is used. 

**Expected Behaviour**
```
"actions": [
      {
        "type": "answer",
        "expression": "Singapore is a place with a population of 3547809. Here is a map: https://www.openstreetmap.org/#map=13/1.2896698812440377/103.85006683126556"
      },
      {
        "type": "anchor",
        "link": "https://www.openstreetmap.org/#map=13/1.2896698812440377/103.85006683126556",
        "text": "Link to Openstreetmap: Singapore"
      },
      {
        "type": "map",
        "latitude": "1.2896698812440377",
        "longitude": "103.85006683126556",
        "zoom": "13"
      }
    ]
```
If response is of this type, then the client should loop through all the actions and display result one after other with delay provided in the response. So, ideally the result of this response should be :
* 1st message : Singapore is a place with a population of 3547809. Here is a map: https://www.openstreetmap.org/#map=13/1.2896698812440377/103.85006683126556
* 2nd message after 0 delay : [Link to Openstreetmap: Singapore](https://www.openstreetmap.org/#map=13/1.2896698812440377/103.85006683126556)
* 3rd message after 0 delay: A map displaying the given coordinates and zoom level 

For response this 
```
"actions": [
      {
        "type": "answer",
        "expression": "Line 1"
      },
      {
        "type": "answer",
        "delay": 400,
        "expression": "Line 2"
      },
      {
        "type": "answer",
        "delay": 1000,
        "expression": "Line 3"
      }
    ]
```
Response should be 
* 1st message Line 1
* 2nd message after 0.4 seconds of 1st message : Line 2
* 3rd message after 1 second of 1st message : Line 3

uday96:
@chiragw15 Cool, I was also wondering about the same. So the delay is only in rendering the response right? 

I will work on this :+1:

chiragw15:
Yes. User should see messages one after another with given delay. Same applies for text to speech. Text1 should be converted to speech. After given delay, text2 and so on.

uday96:
@chiragw15 are there any working queries with which i can test the delay attributes ?

chiragw15:
"multi3"

rishiraj824:
@uday96 is this issue fixed? or needs to be closed?

uday96:
[...]

PR #161, opened 2017-06-06
Fixes 151 - Loop Through Actions and display msgs

uday96:
Fixes issue #151 (partly)

**Changes:**
Instead of combining the data from all action elements and displaying a single response, we loop through each action element and render it.

As of now, every element is displayed with 0 delay.
Delay can be implemented creating a different issue.

**Demo Link:** http://loopactions.surge.sh/

**Sample Queries :**
1. Where is *
2. Search for * 

loopaction

### DIFF
--- a/src/components/ChatApp.css
+++ b/src/components/ChatApp.css
@@ -270,7 +270,7 @@ textarea {
 }
 .leaflet-container {
   height: 150px;
-  width: 80%;
+  width: 250px;
   margin: 0 auto;
 }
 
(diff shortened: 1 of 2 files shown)


uday96:
@rishiraj824 @madhavrathi @isuruAb please review :smile:

src/components/ChatApp.css
```diff
@@ -270,7 +270,7 @@ textarea {
 }
 .leaflet-container {
   height: 150px;
-  width: 80%;
+  width: 250px;
```
isuruAb: this will affect for responsiveness ?

uday96: Initially since text was also displayed in the same div, width auto was working good. Now since its the only element, width: 80% is not working. So i tried with vw also, but 250px seemed fine. 
Can suggest if you know a workaround.

PR #170, opened 2017-06-06
Fixes #162 - Handle coloumn names as array in table action

### LINKED ISSUE #162: Client doesn't handle columns as array from the response
**Actual Behaviour**

When the action type is table, the client expects the column names and parse keys as object as of now. Doesn't handle single column or column nameless response given as array. 

Example response : http://api.susi.ai/susi/chat.json?q=Susi,%20Please%20read%20the%20RSS%20feed%20from%20http://news.ycombinator.com/rss

{
  "type": "table",
  "columns": [
    "title"
  ]
} 

**Expected Behaviour**
Handle such cases and render a table

**Steps to reproduce it**

Query : Susi, Please read the RSS feed from https://news.ycombinator.com/rss

Yes

uday96:
Fixes issue #162 

**Changes:**
Client now handles the case when the coloumns returned in table action type is an array or a key value object.

**Demo Link:** http://tablecolsarray.surge.sh/

**Screenshot:**

tablecol

### DIFF
--- a/src/components/MessageListItem.react.js
+++ b/src/components/MessageListItem.react.js
@@ -98,11 +98,20 @@ function drawWebSearchTiles(tilesData){
 }
 
 function drawTable(coloumns,tableData){
-  let tableheader = Object.keys(coloumns).map((key,i) =>{
+  let parseKeys;
+  let showColName = true;
+  if(coloumns.constructor === Array){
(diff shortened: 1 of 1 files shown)


uday96:
@rishiraj824 @isuruAb @madhavrathi please review :smile:

Issue #173, opened 2017-06-07
create 404 page for Susi web chat


isuruAb:
**Actual Behaviour**

There is no 404 not found page for chat application and it shows default 404 message of Github pages.

**Expected Behaviour**

We should have productive 404 page not found page

**Screenshots of the issue**

<img width="685" alt="screen shot 2017-06-07 at 2 50 57 pm" src="https://user-images.githubusercontent.com/7692626/26871555-3b3e7532-4b91-11e7-8c56-10a1a72c8d41.png">

**Would you like to work on the issue?**

yes

rishiraj824:
@isuruAb Opening the issue as it still does not solve. I think the correct approach to use a 404 page with gh-pages should be to add a 404/html in `public` folder or have a 404.html in gh-pages branch.
http://chat.susi.ai/hey

isuruAb:
@rishiraj824  yes finally i understood the problem. This is an issue with gh-pages .This is happening because users aren't being served the index.html which bootstraps your app when they visit those URLs (404 and other URLs) directly. when we are deploying app in gh-pages we need to add a copy of index file as 404.html page . and if we are deploying it in surge.sh we have to add a copy of index file as 200.html page.

rishiraj824:
@isuruAb Yes, I had talked to you about 404.html. This is really good. Good Work :smile:

PR #181, opened 2017-06-08
Fixes:#173 added 404 page

isuruAb:
Fixes issue #173 

Changes: Added separate page for Not Found page 

Demo Link: surge does not show my 404 page in deployed version.it shows surge 404 page, but in local it is okay.It will work correctly after it merged to master branch.

Screenshot:
screen shot 2017-06-07 at 9 11 27 pm


isuruAb:
@rishiraj824  In my local application it works fine.  are you sure about your method ?

PR #243, opened 2017-06-12
Fixes #234 - seperate messages datewise

### LINKED ISSUE #234: Separate messages datewise
**Actual Behaviour**

Currently, the message history is not separated datewise like in other clients

**Expected Behaviour**

The message date sent, can be shown to separate the messages according to the date sent.

uday96:
Fixes issue #234 

**Changes:**
* The messages are now seperated datewise
* Date is displayed when the message dates are different, thus seperating the messages datewise

**Demo Link:** http://adddate.surge.sh/

**Screenshots:**


uday96:
@rishiraj824 @isuruAb @madhavrathi please review and suggest css for the date section.

src/components/ChatApp/ChatApp.css
```diff
@@ -151,6 +151,18 @@ header{
     padding: 0;
 }
 
+.message-container.Date{
```
rishiraj824: the class message-container is containing a false ::after element which is creating a white arrow section. You can avoid using that class.

src/components/ChatApp/ChatApp.css
```diff
@@ -151,6 +151,18 @@ header{
...
+    margin: 0 auto;
```
rishiraj824: Align the Date, not in center

uday96:
@rishiraj824 made the requested changes.

uday96:
@isuruAb @rishiraj824 aligned date to center also. check the demo link now

Issue #183, opened 2017-06-08
Implement configurations instead of hard-coding


mariobehling:
Currently many configurations in the app are hard-coded, e.g. the app name, design and the server.

If other users want to install the app, they would want to use another name for their app site. Therefore, please define items in a configuration file, so they can be changed if needed.

Include the following:
* URL of SUSI.AI server
* Name of application (e.g. if a user wants to use another name)
* Enable theming (the goal is that users can add themes, that they put in a folder)
* Anything else that can be defined in a configuration file

uday96:
will work on this :+1:

uday96:
@mariobehling does the url of SUSI server mean the url that we are using to fetch responses i.e `api.susi.ai/susi/chat.json` ?

We can also include all the urls used for server calls like
* call to fetch responses : api.susi.ai/susi/chat.json
* call to memory servlet : api.susi.ai/susi/memory.json
* calls for the accounting system like login signup and forgot password
in the config file .

For ex, we recently had to change the server call from api.asksusi.com to api.susi.ai , so including these in config files could prove useful incase the urls change in future

rishiraj824:
The URL configuration has been implemented #215 It means if a user wants to use a custom server of susi, he can fill in the details

uday96:
@rishiraj824 Okay, will add that to user preferences store. Any other additions to the configuration file?

uday96:
@daminisatya @rishiraj824 what else can be included in the configuration file?

uday96:
The theme and server urls have been implemented already in the config or UserPrefernecesStore, I cant think of more inputs to the config file, i'll keep working on other issues and get back to this if I find anything relavant. In the meantime other devs can take this up if they want to.
@rishiraj824 @daminisatya please unassign me for now :smile:

rishiraj824:
closing as all implementations have been done. Will open subissues if more configurations are to be added.

PR #246, opened 2017-06-13
Fixes #192 (Partly) - Save Theme Settings according to user

### LINKED ISSUE #192: Save Theme Setting according to the user
**Actual Behaviour**

This is a sub-issue for #183. There is no theme setting preferred for the user. 

**Expected Behaviour**

One needs to throw the theme setting to the server.

Anyone can take the issue.

uday96:
Fixes issue #192 (Partly)

**Changes:**

  * Added a UserPreferences Store to save all user preferences or defaults.
    This can be further used to add other things and remove hard coded configurations - https://github.com/fossasia/chat.susi.ai/issues/183
 * Added a settings option and dialog, which consists of change default theme option as of now.
   The default theme is updated in the UserPreferenceStore on submit (check console.log)
 * When the page is initially loaded the theme is queried from the default theme in user preferences and set accordingly

Needs implementation of pushing the user preferences store data to server (whenever the user logs out and must be pulled whenever user logs in ?)

**Demo Link:** http://userdefaults-theme.surge.sh/

**Screenshots:** 

settings


uday96:
@rishiraj824 @isuruAb @madhavrathi please review and suggest changes

uday96:
@rishiraj824 yes. We need to have implementation based on user account only. I was implying that we can use this store to store the defaults, remove the hard coded configurations by fetching data from this store and initialising things. When the user logs in we just have to take the user data from the server and set the values in this userpreferences store and when he logs out, we push the userpreferences store data back to server. what do you think?

rishiraj824:
@uday96 Okay. The only thing is it should save preferences when one chooses the theme preferences after logging in, closes the tab and opens it again, and sees that theme there. If that's the case it solves this issue.

uday96:
@rishiraj824 yes, One can only choose preferences after loggin in. But shouldnt the default theme be rendered only when the user logs out and then logs in again? Otherwise he can simply use the change theme option right? Also, we need to push and pull userpreferences store data from the server to test this. That hasnt been implemented as of now. 

The basic idea of this store is :

* Store all the preferences / defaults for the user (which we would store on the server)
* Use these preferences upon initialisation / log-in to initialise the page (also helps in removing the hard coded configuration  -  https://github.com/fossasia/chat.susi.ai/issues/183) 
* Upon login pull user data from server and set the values in the userpreference store
* Upon logout push the data from the userpreference store to server
So the changes in this store can only be viewed after login - logout (when the pulling and pushing to server is also implemented)

mariobehling:
Please ensure theme settings are saved on the server as well (might need to be implemented).

Issue #245, opened 2017-06-13
Change width of Sign In Dialog box 


isuruAb:
**Actual Behaviour**

Width of dialog box is too big.

**Expected Behaviour**

it should be fit to the signIn form.

**Screenshots of the issue**

Where-ever possible attach a screenshot of the issue.
<img width="780" alt="screen shot 2017-06-13 at 8 25 19 am" src="https://user-images.githubusercontent.com/7692626/27064017-9932d6ee-5012-11e7-9ddd-1cb33f0df4e5.png">

**Would you like to work on the issue?**

yes

PR #250, opened 2017-06-13
Fixes: #245 reduce extra width of Dialog-box

isuruAb:
Fixes issue #245 

Changes: reduce extra width of Dialog-box

Screenshots for the change:  
screen shot 2017-06-14 at 2 00 45 am


isuruAb:
@rishiraj824  this happens because Login form has a static width. I will make it responsive in another issue.

mariobehling:
Did you try this out? Which alternative sample server did you use? There is no sign up for an alternative server. Does this mean the whole wep-app uses another server? Does it work?

PR #262, opened 2017-06-15
Fixes #253 - Show table content on hover

### LINKED ISSUE #253: The full content present in a column, not shown in table type output.
**Actual Behaviour**
When we hover over a name in a column in table type output, the full content of that column is not shown.

**Expected Behaviour**
The content on that column should be shown into a text box above it like this:
screenshot 236

**Screenshots of the issue**
In this screenshot, if we hover over an author name in this table its full content is not shown.
webchathovertitle

@isuruAb, would you like to work on it?

uday96:
Fixes issue #253 

**Changes:**
* Added `<abbr> tag` to elements in table to show the entire cell content on hover

**Demo Link:**  http://tablehover.surge.sh/

table

### DIFF
--- a/src/components/ChatApp/MessageListItem.react.js
+++ b/src/components/ChatApp/MessageListItem.react.js
@@ -139,7 +139,9 @@ function drawTable(coloumns,tableData,count){
       return(
         <TableRowColumn key={i}>
           <Linkify properties={{target:'_blank'}}>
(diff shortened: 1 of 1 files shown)


uday96:
@rishiraj824 @isuruAb @madhavrathi @daminisatya please review

Issue #221, opened 2017-06-11
Can't go to SignUp page using URL


isuruAb:
**Actual Behaviour**

When we navigate to signup page by clicking top right corner button it works correctly.but when we click on this link directly it moves to 404 page http://chat.susi.ai/signup

**Expected Behaviour**

User should have ability to move to signup page directly by clicking this url.
http://chat.susi.ai/signup

**Would you like to work on the issue?**

yes

rishiraj824:
@isuruAb How are you proceeding with this? As far as I know <Link to="/" /> can be only used to navigate between the links while using `react-router`. Given that this is a static single page.

PR #274, opened 2017-06-16
Fixes: #221 can go to urls directly

isuruAb:
Fixes issue #221 
Changes: now users can go to routes directly


mariobehling:
Very good.

Issue #282, opened 2017-06-18
Custom theme options should align center 


isuruAb:
**Actual Behaviour**

All the options are right aligned .
custom theme dialog box has extra width.

**Expected Behaviour**

should be center aligned and  width should be reduced .

**Screenshots of the issue**

<img width="586" alt="screen shot 2017-06-18 at 5 34 34 am" src="https://user-images.githubusercontent.com/7692626/27257213-8210dc22-53eb-11e7-9365-ae2d5646a06c.png">

**Would you like to work on the issue?**

yes

PR #287, opened 2017-06-19
fixes:#282 change width of dialog box and align center

isuruAb:
Fixes issue #282 

Changes: Align dialog box content to center. and responsive

Screenshots for the change: 
screen shot 2017-06-18 at 1 43 02 pm

### DIFF
--- a/src/components/ChatApp/ChatApp.css
+++ b/src/components/ChatApp/ChatApp.css
@@ -461,3 +461,8 @@ a {
 em{
   background-color: orange;
 }
+.settingsComponents{
+  width: 90%;
+  text-align: center;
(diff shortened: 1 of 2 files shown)


mariobehling:
@isuruAb @daminisatya I pointed out a number of times, that certain design choices must be stored in the settings of the server. We have different clients and we need to align these clients. Therefore any changes to the design need to be stored in the settings - with the server. Are you aware of that? I see that there are implementations of design, but just on the web client

Regarding merges: We should distribute responsibilities more and more. In usual cases two contributors should review PRs before a third one merges it. This also helps to keep everyone aligned.

Issue #279, opened 2017-06-17
Improve usage of screen space of top bar on mobile view


mariobehling:
The top bar on mobile devices is very big. What is the best option here? Please start with reducing the height of the top "SUSI" bar on mobile devices.

isuruAb:
working on it

isuruAb:
@mariobehling  this can be closed

rishiraj824:
closing via #290

PR #290, opened 2017-06-19
Fixes:279 reduce app-bar size 

isuruAb:
Fixes issue #279 
Changes: reduce height of the app-bar  on search state and default satate


isuruAb:
@rishiraj824  when you got this problem ?

Issue #233, opened 2017-06-12
Feature to add a custom theme


rishiraj824:
**Actual Behaviour**

Currently, one can select a pre-defined theme but cannot create a custom theme

**Expected Behaviour**

One should be able to set a custom color in the Chat.
One should also be able to save a theme https://github.com/fossasia/chat.susi.ai/issues/192

**Would you like to work on the issue?**

Anyone is free to take this up.

uday96:
So creating a custom theme as in the user must be given an option to choose styling for everything?

The header background
Font color
document background etc?

And must also be given an option to preview a static page or so with the selected option?
And then when he is satisfied, click on submit to store that as a custom theme?

rishiraj824:
@uday96 Yes, all the changes which are currently happening while switching themes. 
Maybe a dialog can be used to let users choose the color for each segment. For a preview, the user can see the changes directly as the color is changed.

rishiraj824:
@uday96 I can work on this partly 
- To create a dialog for changing themes.
- Preview the theme directly.

uday96:
@rishiraj824 sure :+1:

PR #296, opened 2017-06-19
Fixes #295 - Remove Change Theme as drop down and include it in settings

### LINKED ISSUE #295: Remove Change Theme as an drop down option and include in settings
**Actual Behaviour**

Change Theme option is a drop down option as of now. And the settings include default theme option. 
The default theme option in the settings was originally idealised as the default theme to render when the user logs in irrespective of the theme when the user previously logged out.

**Expected Behaviour**

We must save the theme in userpreferences store i.e a part of user data which is intended to push and pull from server (once server side support is implemented) everytime user changes theme so that when he logs in, the same theme is rendered which was used when he previously logged out.

Yes

uday96:
Fixes issue #295 

**Changes:**
* Removed theme from SettingsStore and all the actions related to it
* Used theme in UserPreferencesStore to store theme upon every theme change action

**Needs to be implemented:**
* Pushing and pulling the user data (UserPreferencesStore data) to and from server so that the changes  are maintained upon refreshing the page or login logout.

**Demo Link:**  http://changethemesetting.surge.sh/

**Screenshots:**

settings

### DIFF
--- a/src/actions/Settings.actions.js
+++ b/src/actions/Settings.actions.js
@@ -17,13 +17,6 @@ export function setDefaults() {
   });
 }
 
-export function setDefaultTheme(defaultTheme){
-  ChatAppDispatcher.dispatch({
-    type: ActionTypes.DEFAULT_THEME_CHANGED,
-    defaultTheme
-  });
-}
-
 export function setDefaultServer(defaultServer){
   ChatAppDispatcher.dispatch({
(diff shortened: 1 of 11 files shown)


uday96:

Have put `Custom Theme` option as a drop down in logged in mode for now. Can change it accordingly when working on https://github.com/fossasia/chat.susi.ai/issues/233 and also clarify it once as themes need to be accessible by all clients as mario mentioned.

PR #332, opened 2017-06-26
Fixes #285 - Search Field Animated and App Bar Fixed

### LINKED ISSUE #285: Improve and fix top bar related to search box
The top bar changes height when the search field is activated.

Please also implement an expansion of the search box to the left and when it is expanded move "SUSI" to the left. 

Similar to this effect here: https://www.w3schools.com/howto/howto_css_animated_search.asp
screenshot from 2017-06-18 13-50-21

uday96:
Fixes issue #285 

**Changes:**
* Search field is now animated to expand upon focus or hover
* Moved search section into message section
* In mobile screens, due to space constraints, only the search filed will be shown and the SUSI logo disappears

**Needs to be implemented**
* Move case sensitive option for search from search section into this
* Delete search section if this is merged and case sensitive option is also moved

Will implement the above once this is accepted and merged

**Demo Link:** http://searchanimation.surge.sh/

**Screencast:**

searchanim


uday96:
@mariobehling @rishiraj824 @madhavrathi @isuruAb Please review and suggest changes.

uday96:
@madhavrathi fixed it, please check again

uday96:
@mariobehling If we remove the -> arrow and close search when users click outside the search field, what if the user wants to maually scroll the chat area (though buttons are there)? Also in mobile phones, mostly user would swipe so that would exit the search field in this case. The browser search box also has a close or `x` button. Should i add that instead?

mariobehling:
@uday96 Ok, please propose a better solution that works well on mobile. The current implementation in this UI does not feel natural. How are other mobile apps are solving this?

mariobehling:
The implemented effect is also confusing. Actually instead of having one effect, there are two effects.
1. You activate the fields.
2. You expand the fields.
3. On scroll over you expand the search box even more.

Expected: Only show fields as you expand. And show any UI buttons/elements on the right hand side of the search box.

uday96:
@mariobehling The existing apps like whatsapp also use a exit or back button along with the up down buttons.So does the browser search. The current implementation of search in chat.susi.ai is similar.

The animation for the search field is only possible on focus or hover. So the field is autofocused to give the initial expanding animation. Later it collapses or expands based on hover or focus of the field.
Moving the UI buttons to right will look like  - http://search123.surge.sh/

madhavrathi:
@uday96 I think you're right, whatsapp is having similar implementation, Buttons to the right and the rest is all same.

mariobehling:
Use only expansion effect for the whole area including the up and down arrow, please.

uday96:
@mariobehling Please check out : http://searchanimation.surge.sh/ . Does this look fine? We can use either the search icon again to close search mode or the close icon. Which one to keep or remove?
Also any other changes?

mariobehling:
Good job! Thank you.

PR #336, opened 2017-06-27
Fixes #180 - Get error message from ajax error() when server throws error

### LINKED ISSUE #180: Get error message from ajax error() when server throws API exception
**Actual Behaviour**

When we make a server call for password recovery and it doesnt succeed, the server throws a API exception with error code and status text. But the same when queried through ajax.error() func, we always get status 404 and 'error'

**Expected Behaviour**

Get the error status and msg sent by the server

**Steps to reproduce it**

http://api.susi.ai/aaa/recoverpassword.json?forgotemail=

Opening the url directly in browser gives a status code 422 while the same when queried through ajax gives 404

**Screenshots of the issue**

err

Anyone can take this up

rishiraj824:
Fixes issue #180

Changes:
- Currently handling all Ajax errors
- Showing appropriate messages according to status codes.

Demo Link: 
http://susi.surge.sh

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,6 +14,7 @@
     }
   },
   "env": {
+    "browser": true,
     "jasmine": true
   },
   "globals": {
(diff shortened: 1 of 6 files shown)


rishiraj824:
@uday96 sure

rishiraj824:
@uday96 Please check now.

PR #372, opened 2017-07-01
Fixes #370 - Added EnterAsSend Setting

### LINKED ISSUE #370: Add setting to choose between sending query on `Enter`
**Actual Behaviour**

The messages are sent upon pressing `Enter` key.

**Expected Behaviour**

Give the user an option to choose `Enter` key as send or not like in other chat clients

enter

Yes

uday96:
Fixes issue #370 

**Changes:**
* Added EnterAsSend setting where user can choose to send query upon pressing enter or not
* If query is not sent on ENTER, then on clicking send button, '\n' is replaced with '  ' and sent similar to android client.

**Demo Link:** http://enterassend.surge.sh/

enter


uday96:
@rishiraj824 @madhavrathi @isuruAb please review

Issue #364, opened 2017-06-30
search fields is not aligned correctly in mobile view


isuruAb:
**Actual Behaviour**

search fields is not aligned correctly in mobile view

**Expected Behaviour**

search field should be correctly aligned and  should function correctly.

**Screenshots of the issue**

![wp_ss_20170701_0001](https://user-images.githubusercontent.com/7692626/27756129-96bd4f12-5e12-11e7-835c-6c48879e65d3.jpg)

**Would you like to work on the issue?**

Please let us know if you can work on it or the issue should be assigned to someone else.
yes

PR #373, opened 2017-07-03
Fixes: #364 search field aligned correctly on mobile view

isuruAb:
Fixes issue #364 

Changes:  reduce width of the search field and gave a percentage of width.

Demo Link: http://isuruab1.surge.sh/


uday96:
@isuruAb For <300 px its showing blank without any option to go back too. Is it not possible to fit the search section for lesser widths? If not maybe we should at least give an option to go back

Issue #405, opened 2017-07-06
Change send button icon when not using mic


madhavrathi:
**Actual Behaviour**

Mic icon does not appears after sendind a message.

**Expected Behaviour**

It should appear back.

**Would you like to work on the issue?**

Yes

rishiraj824:
closing via #406

PR #406, opened 2017-07-06
Fixes#405: Send button icon changes properly

madhavrathi:
Fixes issue #405 

Changes: Now send button icon changes back to mic after sending message.

Demo Link: [Add here the link where your changes can be seen]

Screenshots for the change:http://susimadhav.surge.sh/
 
issuefix
@uday96 @rishiraj824 @isuruAb @daminisatya Please Review.
Thanks

### DIFF
--- a/src/components/ChatApp/MessageComposer.react.js
+++ b/src/components/ChatApp/MessageComposer.react.js
@@ -22,7 +22,7 @@ const loadingStyle = {
     fill: '#607d8b',
     marginTop: '10px'
 }
-let Button = <Mic />, speechRecog;
+let Button, speechRecog;
 // Getting the Speech Recognition to test whether possible
 const SpeechRecognition = window.SpeechRecognition
       || window.webkitSpeechRecognition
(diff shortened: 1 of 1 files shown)


src/components/ChatApp/MessageComposer.react.js
```diff
@@ -22,7 +22,7 @@ const loadingStyle = {
     fill: '#607d8b',
     marginTop: '10px'
 }
-let Button, speechRecog;
+let Button = <Mic />, speechRecog;
```
rishiraj824: We cannot have the Button to be `<Mic />` by default as it first needs to be checked whether the Browser supports the Speech API. So it is just initialized that way. Also, I am not able to produce this issue in http://chat.susi.ai Can you please check once? @madhavrathi

src/components/ChatApp/MessageComposer.react.js
```diff
@@ -164,6 +164,7 @@ class MessageComposer extends Component {
           Actions.createMessage(text, this.props.threadID);
         }
         this.setState({text: ''});
+        Button = <Mic />
```
rishiraj824: The same reason applies here, we need a check to confirm that the browser follows the Speech Recognition API first. If it doesn't, then the user will be stuck with a Mic infinitely.

madhavrathi:
@rishiraj824 Please review again 
Thanks

madhavrathi:
@rishiraj824 There was a delay in updating the deploy link. Please check again http://susimadhav.surge.sh/
Thanks

Issue #412, opened 2017-07-07
Change background for message section


madhavrathi:
**Actual Behaviour**

Change background for background is already being implemented. 

**Expected Behaviour**

We should have an option to change background of message section as well

**Would you like to work on the issue?**

Yes

rishiraj824:
@madhavrathi @isuruAb I think it should actually be implemented for the Message Section only and not for the whole body. What do you suggest?

madhavrathi:
@rishiraj824 For Web user can choose background for whole body as well, for mobile only message section. I guess there is no harm in having both features.

rishiraj824:
@madhavrathi Okay, let's try that out then. No harm in trying it out. I think the wallapaper functionality is also there in other clients, if I am not wrong?

rishiraj824:
closing via #415

PR #415, opened 2017-07-07
Fixes#412: user can now add background image to message section

madhavrathi:
Fixes issue #412 

Changes: Now user can choose background for message section as well.

Demo Link: http://susimadhav.surge.sh/

Screenshots for the change: 

aajkaissuefix
@rishiraj824 @uday96 @isuruAb @daminisatya Please Review
Thanks

### DIFF
--- a/src/components/ChatApp/MessageSection/MessageSection.react.js
+++ b/src/components/ChatApp/MessageSection/MessageSection.react.js
@@ -38,6 +38,7 @@ function getStateFromStores() {
     composer:'',
     body:'',
     bodyBackgroundImage:'',
+    messageBackgroundImage:'',
     searchState: {
       markedMsgs: [],
       markedIDs: [],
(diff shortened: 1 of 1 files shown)


src/components/ChatApp/MessageSection/MessageSection.react.js
```diff
@@ -546,7 +565,7 @@ class MessageSection extends Component {
                 <ul
                   className='message-list'
                   ref={(c) => { this.messageList = c; }}
-                  style={{background:pane}}>
+                  style={messageBackgroundStyles}>
```
rishiraj824: Let's have the `background-repeat:none` property as well to the styles. @madhavrathi

madhavrathi:
@rishiraj824 Please Review. @isuruAb I am able to change image just fine, can you try again with a different image. For ex. try this url https://image.freepik.com/free-vector/abstract-blue-background_1048-1511.jpg
Thanks

rishiraj824:
@uday96 Probably, this will only be a one time use feature because currently, the server is not going to store any file of a user. If I am not wrong the Android Client does it but it happens locally only on the App https://github.com/fossasia/susi_android/pull/504

rishiraj824:
screenshot from 2017-07-08 14 32 47
@madhavrathi What I meant by adding a `background-repeat : none` was that the background should not repeat itself over the area and cover the whole area. So maybe https://www.w3schools.com/cssref/playit.asp?filename=playcss_background-size&preval=cover
this link can help

madhavrathi:
aajkaissuefix

@rishiraj824 It all looks good now. Please Review http://susimadhav.surge.sh/

PR #450, opened 2017-07-11
Fixes #445 - Support to convert HTML strings to React elements

### LINKED ISSUE #445: HTML parser support for text
**Actual Behaviour**

HTML content as strings is not automatically parsed into HTML elements

**Expected Behaviour**

Parse strings to convert into respective HTML entities if any

**Screenshots of the issue**

dream pad link : http://dream.susi.ai/p/anchortest
query : html text

ht

Yes

uday96:
Fixes issue #445 

**Changes:**
* Added a new library [html-react-parser](https://www.npmjs.com/package/html-react-parser) to convert HTML strings into React elements and also handle Special HTML entities

**Demo Link:** http://htmlsupport.surge.sh/

**Etherpad Link** http://dream.susi.ai/p/anchortest

**Sample Queries**:
-  html demo
- html text

htmldem

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "classnames": "^2.2.5",
     "flux": "^3.1.2",
     "history": "^4.6.1",
-    "html-entities": "^1.2.1",
+    "html-react-parser": "^0.3.5",
     "jquery": "^3.2.1",
     "keymirror": "^0.1.1",
     "leaflet": "^1.0.3",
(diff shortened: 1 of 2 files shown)


uday96:
@rishiraj824 @isuruAb @madhavrathi @daminisatya please review

rishiraj824:
@uday96 What is the difference between parser and htmlEntities which was used previously?

uday96:
@rishiraj824 `htmlEntities` only parsed for special html chars like `&lt` or `&#38;` but if the string contains HTML tags like `<b>hey</b>' or anything else, then they weren't parsed. This `Parser` handles both the cases - special characters and HTML tags. Generally it should be automatically handled for regular HTML rendering, but since react uses JSX, we are handling this seperately

Issue #383, opened 2017-07-04
Propose a way to show the SUSI skills on first appearance


rishiraj824:
**Actual Behaviour**

There is no introductory message or process to show SUSI Skills or actions when one first opens the Website or logs in as mentioned by @chiragw15 in https://github.com/fossasia/susi_android/issues/785.

**Expected Behaviour**

Have an introductory pane or showcase skills like Cortana to show what important actions or skills SUSI has. One needs to also keep in mind that this appears only once, i.e. only the first time the User open's the site or logins.

One can also have this as an option somewhere in the Chat App to see the skills or actions when he wants. But showing this at the first appearance should be mandatory.

madhavrathi:
Will try this. I will propose an initial UI and let's see what other's think of it.

uday96:
@rishiraj824 @madhavrathi I think we need to first confirm with @mariobehling regarding this.

madhavrathi:
@mariobehling What are your thoughts on this. Please comment

mariobehling:
Implement this as a skill!

https://github.com/fossasia/chat.susi.ai/pull/453#issuecomment-314682006

PR #453, opened 2017-07-11
Added a dialog to showcase susi skills

madhavrathi:
Fixes issue #383 (Partially)
Changes: Added a dialog to showcase susi skills. Visual enhancements will be done in future PR. This issue will get many PR's as it is a big issue.
next steps:
1) Add custom images to skills showcase.
2) Modify the position of Dashboard buttons.
 further issues will be created.
Demo Link: http://susi-skillshow.surge.sh/

Screenshots for the change: 
aajkaissue4
@rishiraj824 @uday96 @isuruAb @chiragw15  Please Review
Thanks

### DIFF
--- a/src/components/ChatApp/ChatApp.css
+++ b/src/components/ChatApp/ChatApp.css
@@ -434,9 +434,10 @@ textarea {
   -webkit-box-shadow: none;
   -moz-box-shadow: none;
   box-shadow: none;
-  width: 92%;
+  width: 85%;
   resize: none;
   border-radius: 6px;
+  margin-left: 40px;
   padding-left: 10px;
   padding-top: 10px;
   font-size: 14px;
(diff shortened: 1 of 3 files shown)


uday96:
@madhavrathi I dont think hard coding skills is a good idea. We have endpoints to get sample queries from the server and also filter queries based on models and experts. I think we need to first figure things out with other clients too as to how they are planning, because incase they are hardcoding too, then we need to hardcode the same skills.

madhavrathi:
@uday96 I discussed it with @chiragw15. Skills display needs to be displayed and I guess it can be hard coded in respective platforms. Please suggest otherwise.
Thanks

mariobehling:
This should be solved as a skill not in the UI. The idea behind SUSI is to make an assistant that you can talk to. We don't need more buttons. Ask SUSI "What can you do" and make it into a skill!

Issue #462, opened 2017-07-12
Improve Sign up Popup


mariobehling:
Improve as follows:

* Speed up effect where popup slides in from top as much as possible
* Avoid internal scrollbar
* Take out "Standard Server" and only keep custom server. Make the custom server button into a button that can be activate or deactivated. When it is activated it should show URL field in the place of the the text "Custom Server"
* Design the area more space savy similar to the Facebook login

isuruAb:
Will work on this

isuruAb:
@uday96  @rishiraj824  @madhavrathi  guys please give your suggestions to Avoid internal scrollbars 
 I'm hoping to take the signup page to another new route like /signup then we can build a seperate view for signup.

madhavrathi:
@isuruAb this is a good option. Please confirm with @mariobehling before proceeding.

PR #483, opened 2017-07-14
Fixes: #462 (partially) Add checkbox instead of radio button group

isuruAb:
Fixes issue #462 Partially

Changes: Remove radio button group and replace it with checkbox

Screenshots for the change: 

screen shot 2017-07-19 at 8 42 24 am

@rishiraj824  @madhavrathi  @uday96  please review this

### DIFF
--- a/src/components/Auth/SignUp/SignUp.react.js
+++ b/src/components/Auth/SignUp/SignUp.react.js
@@ -304,6 +304,7 @@ export default class SignUp extends Component {
                             name="serverUrl"
                             className="serverUrl"
                             onChange={this.handleServeChange}
+                            onTouchTap={this.handleServeChange}
                             value={this.state.serverUrl}
                             errorText={this.customServerMessage}
                             floatingLabelText="Custom URL"
--- a/src/components/Auth/SignUp/SignUp.react.js
+++ b/src/components/Auth/SignUp/SignUp.react.js
@@ -76,12 +76,43 @@ export default class SignUp extends Component {
             });
         }
     }
+    handleServeChange=(event)=>{
+        let state = this.state;
+        let serverUrl
+        if (event.target.value === 'customServer') {
+            state.checked?state.checked=false:state.checked=true;
+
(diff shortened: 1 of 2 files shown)


uday96:
@isuruAb This looks neat :+1: But @mariobehling mentioned that : 
>Take out "Standard Server" and only keep custom server. Make the custom server button into a button that can be activate or deactivated. When it is activated it should show URL field in the place of the the text "Custom Server"

Please confirm once as to which method should be followed

isuruAb:
@uday96  can you check this again

mariobehling:
>  When it is activated it should show URL field in the place of the the text "Custom Server"

When the field is activated in the place of "custom server" show the field to add the custom URL!

uday96:
su

@isuruAb There is a slight overlap between error text and sign up button. Please check. Rest looks good :+1:

mariobehling:
Open the signup page and click on the movable button. Result in my view: There is a "jump", that should not be there.

Please increase the distance between "custom server" and "sign up" to be more natural.
screenshot from 2017-07-19 08-21-24

mariobehling:
Could you make the field that shows up where to enter the "URL" exactly in the same position as "Use Custom Server" please?

src/components/Auth/SignUp/SignUp.react.js
```diff
@@ -280,24 +295,22 @@ export default class SignUp extends Component {
...
     render() {
-
-        const serverURL = <TextField name="serverUrl"
-                            onChange={this.handleChange}
+        const customUrlStyle= {
+            width:'202px',
+            textAlign:'left',
+            margin:'-35px 0 0px 5px',
+        }
+        const serverURL = <TextField
```
rishiraj824: @isuruAb Add a onClick function and handle the this.state.checked state there to achieve the effect.

mariobehling:
Good job! Would be great, if you could also do this for the login and settings.

PR #498, opened 2017-07-16
Fixes #492 - implement text-to-speech with correct language

### LINKED ISSUE #492: implement text-to-speech with correct language
please implement the evaluation of the "language" key as described in https://github.com/fossasia/susi_android/issues/853

rishiraj824:
Fixes issue #492 

Changes: 
- Now getting voice outputs based on the lang property in message
```
{
  "query": "hello",
  "count": 1,
  "client_id": "aG9zdF84NC4xODEuMTEyLjEzOA==",
  "query_date": "2017-07-15T21:31:09.833Z",
  "answers": [{
    "data": [{
      "0": "hello",
      "1": "",
      "token_original": "hello",
      "token_canonical": "hello",
      "token_categorized": "hello",
      "timezoneOffset": "-120",
      "language": "en"
    }],
    "metadata": {"count": 1},
    "actions": [{
      "type": "answer",
      "language": "en",
      "expression": "Hello!"
    }],
    "skills": ["/susi_server/conf/susi/en_0700_ai_play.json"]
  }],
  "answer_date": "2017-07-15T21:31:09.855Z",
  "answer_time": 22,
  "language": "en",
  "session": {"identity": {
    "type": "host",
    "name": "xx.xx.xx.xx",
    "anonymous": true
  }}
}
```

Demo Link: 
http://susi-voice.surge.sh

### DIFF
--- a/src/actions/API.actions.js
+++ b/src/actions/API.actions.js
@@ -30,8 +30,8 @@ export function getLocation(){
       }
     }
   });
-}
 
+}
 export function createSUSIMessage(createdMessage, currentThreadID, voice) {
   var timestamp = Date.now();
   let receivedMessage = {
@@ -45,7 +45,8 @@ export function createSUSIMessage(createdMessage, currentThreadID, voice) {
     date: new Date(timestamp),
     isRead: true,
     type: 'message',
-    voice: voice
+    voice: voice,
+    lang: 'en-US'
     };
 
   let defaults = UserPreferencesStore.getPreferences();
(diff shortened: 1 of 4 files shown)


src/actions/API.actions.js
```diff
@@ -95,6 +96,8 @@ export function createSUSIMessage(createdMessage, currentThreadID, voice) {
       Actions.sendToHardwareDevice(response);
 
       receivedMessage.text = response.answers[0].actions[0].expression;
+      // Setting Language received from User
+      receivedMessage.lang = response.answers[0].actions[0].language;
```
uday96: Add a check to see if response.answers[0].actions[0] contains language attribute or not. Rest looks good 👍

rishiraj824: I think the query model has this by default. If you check, all responses from server have this attribute compulsorily as @Orbiter mentioned in https://github.com/fossasia/susi_android/issues/853
Otherwise following best practices, we even need to have a check for if the `response.answers[0].actions[0]` contains `type` and `expression` which would be redundant because this is the basic query model being followed on the server.

http://api.susi.ai/susi/chat.json?q=hello

uday96: > The action may still have no "language" object assigned, in that case still use the default language of the device.

Incase language attribute is not present like @Orbiter mentioned in https://github.com/fossasia/susi_android/issues/853, then it might throw an exception na?
But yea, like you said if its surely gonna be there then its redundant to add the check 👍

rishiraj824: Yes, agreed, in that case will add the lang as the `browser.locale`

PR #538, opened 2017-07-19
Fixes #475 - Created a new component for custom server

### LINKED ISSUE #475: Create a new component for custom server
**Actual Behaviour**

As of now we are using radio buttons to choose between standard and custom server.
The same is used in login, signup, forgot password and settings. So its better to have a seperate component for the same

**Expected Behaviour**

Take out "Standard Server" and only keep custom server. Make the custom server button into a button that can be activate or deactivated. When it is activated it should show URL field in the place of the the text "Custom Server"

Anyone can take this up

uday96:
Fixes issue #475 

**Changes:**
* Added a new component for custom server
* Used this component in Login, SignUp and ForgotPassword

login

signup

forgotpwd

### DIFF
--- a/src/components/Auth/ForgotPassword/ForgotPassword.css
+++ b/src/components/Auth/ForgotPassword/ForgotPassword.css
@@ -2,14 +2,6 @@
 	margin: 0 auto;
 }
 
-.serverUrl{
- 	float: left;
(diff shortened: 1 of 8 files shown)


uday96:

PR #571, opened 2017-07-22
Fixes #569 - Break word for message text

### LINKED ISSUE #569: Scroll bar shows up if text is longer than width of window instead of breaking text
The answer of a text that is using more width than the window is breaking as expected. Please also break the line of the question. Ensure there is not horizontal scroll bar.

screenshot from 2017-07-22 01-07-56

uday96:
Fixes issue #569 

**Changes:**
* Break line to prevent overflow

overflow


uday96:

PR #608, opened 2017-07-26
Fixes #579 - Dont show I found this on the web for empty websearch rss response

### LINKED ISSUE #579: "I found this on the web" should not be shown
**Actual Behaviour**
With "no results found" reply, "I found this on the web" is also shown. 

**Expected Behaviour**
No need of "I found this on the web" in that case.

**Screenshots of the issue**
noresultsfound

@uday96 @rishiraj824 Do you think the same? If yes, can you work on it. Thanks!

uday96:
Fixes issue #579 

**Changes:**
* Removed `I found this on the web:` when no results are found for websearch

**Demo Link:** http://udaytheja1.surge.sh/

noresults

### DIFF
--- a/src/components/ChatApp/MessageListItem/MessageListItem.react.js
+++ b/src/components/ChatApp/MessageListItem/MessageListItem.react.js
@@ -111,6 +111,7 @@ class MessageListItem extends React.Component {
         if(mapIndex>-1){
           actions = ['map'];
         }
+        let noResultsFound = false;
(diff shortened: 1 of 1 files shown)


uday96:
@rishiraj824 @isuruAb @madhavrathi please review

tabesin:
Coding style can be improved e.g.
> if(websearchTiles.length === 0){
>                 noResultsFound = true;
>               }
but okay.

PR #625, opened 2017-07-27
Fixes #613 , #635 - Added Language settings for TTS and united settings on route /settings

### LINKED ISSUE #613: Add settings to choose default language for TTS
**Actual Behaviour**

There is no option to choose default language for TTS like in other chat clients

**Expected Behaviour**

Add settings to choose default language for TTS

whatsapp image 2017-07-24 at 13 32 11

Yes

uday96:
Fixes issue #613 #635 

**Changes:**
* Query the languages supported by browser for speech synthesis and populates language list in TTS Settings
* If the response from server doesnt have a lang attribute then the default language for TTS will be that selected by user in TTS Settings.
* Stored the TTS lang in server for logged in user

**Needs to be implemented:**
* On clicking `Play a short demonstration of speech synthesis` , The text passed to the Voice Player must also be of the same langugage chosen for proper speech output in that languge.

@rishiraj824 @isuruAb @madhavrathi So I still need to translate the text `This is an example of speech synthesis` to the chosen language and pass it to Voice Player for playing the example. Kindly suggest if you have any ideas.

ttslang


uday96:
@rishiraj824 @isuruAb @madhavrathi Please review. 

As of now, though we choose a different language, when we click on `Play a short demonstration of speech synthesis` , the text passed is in English so we can only find the difference in speech output to be of different accent. But we actually need to pass the translated text to that respective language to play a proper example. Eg: Incase I choose german, then the text passed must be `Dies ist ein Beispiel für die Sprachsynthese` instead of `this is an example of speech synthesis` or say I choose hindi then text passed must be `यह अभिव्यक्ति संश्लेषण के स्प्रेशिनटेशमप्ले के लिए एक एक्स्टियंस है`. And we need this translation only for https://github.com/fossasia/chat.susi.ai/pull/625/files#diff-27414f33628667eb909bf75a16b02f54R25 to play the speech output example. The language output for server response works fine

Will try to fix this in a seperate issue. Please suggest if you have any ideas too.

rishiraj824:
@uday96 This should work, for now, just to send what language is being set by the user. For the demonstration, I think we can figure that out in a different issue. Please confirm with @mariobehling once about this too.

madhavrathi:
@uday96 @rishiraj824 I agree, this should work for now

uday96:
**Changes added :**
 - Shifted settings to new route `/settings`
 - Moved `Change Password` to settings
 - Removed `Chat Anonymously`
 - Replaced Radio Buttons for custom server with the toggle custom server component similar to others

@rishiraj824 @isuruAb @madhavrathi @mariobehling Please review.  For now on submitting settings, the chat page is reloading. If its not fine, will try to prevent the reload in a different issue as it was giving some errors which needs thorough check through the

uday96:
@mariobehling Should I also add a top bar to settings page?

mariobehling:
> Should I also add a top bar to settings page?

@uday96 Settings page should be same on chat page itself.

uday96:
@mariobehling But you asked to move settings to a new route '/settings` and the chat page is at '/' . I can make the settings page identical to chat page by adding a top bar.

uday96:
@mariobehling Please check now. I have added `overview ` to drop down in logged in state too

mariobehling:
* Logo of settings page is not in the same position as on chat page. Should we anyways consider moving the logo to the left on all pages? Maybe do this and position it in the same way as on the Overview page
* Dropdown is not same as on chat page
* There is an arrow on the left that should not be there

uday96:
@mariobehling Please check now.

[ x ] Logo of settings page is not in the same position as on chat page. Should we anyways consider moving the logo to the left on all pages? Maybe do this and position it in the same way as on the Overview page
[ x ] Dropdown is not same as on chat page
[ x ] There is an arrow on the left that should not be there
[ x ] Overview page etc. only show "Logout" as Dropdown. Please follow facebook and ensure the dropdown menu works as expected on all pages.

uday96:
@mariobehling I have disabled the settings if its not supported by the browser now. The languages supported by the browser is queried and the language drop down is populated in the settings.
The sliders for rate and pitch are to choose rate and pitch values and Reset buttons will reset to the default values. Listen to an example will play the text `This is an example of speech synthesis` with the selected language + rate + pitch giving the users a demo of how the speech output will look like. But the language setting is for the default speech output language setting and is used only when the response from server doesnt mention which language to use.

It is working for me. Please check https://vimeo.com/228148465 and try out http://udaytheja.surge.sh/

mariobehling:
> I have disabled the settings if its not supported by the browser now. 

What is the user supposed to do in the UI?I have the latest Chrome browser?

@uday96 The video shows good progres, but this cannot be the end.

I don't see this working on the chat UI. You are only showing the page. Why do you start with advanced options in settings? Why do we have to spend time on settings now? We need this to work on the chat. The language should be choosen according to browser settings first. This can be overriden in the chat settings then.

First step: Show the speech output working on the web chat UI.

uday96:
@mariobehling Now the examples played are in their local languages. Also the flow now is : 
If server response for message doesnt contain language then check for browser lang and if browser lang is also undefined then use the language setting choosed from chat UI.

Please check

mariobehling:
Thank you! Looks good. Let's test this in more detail over upcoming days. Cheers.

PR #627, opened 2017-07-28
Fixes #605 - show feedback thumbs only once for multi action responses

### LINKED ISSUE #605: Show feedback thumbs only once for multi action responses
**Actual Behaviour**

When the response contains multiple actions, multiple chat bubbles are rendered and each chat bubble has a feedback option. But all chat bubbles correspond to the same skill and so feedback must be collected only once to else user can choose different feedback for different chat bubble but the client has to make only a single feedback call giving inconsistency

**Expected Behaviour**

Show feedback thumbs preferably only for the last chat bubble..

**Screenshots of the issue**

thumb

Yes

uday96:
Fixes issue #605 

**Changes:**
*  Show feedback thumbs only for the last bubble of susi response as all the bubbles correspond to the same skill so we must collect response only once

**Demo Link:**  http://susifeedback.surge.sh/

rss

### DIFF
--- a/src/components/ChatApp/MessageListItem/MessageListItem.react.js
+++ b/src/components/ChatApp/MessageListItem/MessageListItem.react.js
@@ -112,6 +112,7 @@ class MessageListItem extends React.Component {
           actions = ['map'];
         }
         let noResultsFound = false;
+        let lastAction = actions[actions.length - 1];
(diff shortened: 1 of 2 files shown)


uday96:
@rishiraj824 @isuruAb @madhavrathi please review

Issue #650, opened 2017-07-31
Logged in users get the login option on static pages


isuruAb:
**Actual Behaviour**

Logged in users get the login option on static pages

**Expected Behaviour**

Logged in users should not get the login option on static pages

**Would you like to work on the issue?**

yes

PR #651, opened 2017-07-31
Fixes: #650 - Logged in users get the login and signup option instead of logout

isuruAb:
Fixes issue #650

Changes:  Logged in users get logout option instead of login and signup options

Demo Link: http://isuruab2.surge.sh/

Screenshots for the change: 
screen shot 2017-07-31 at 9 35 10 pm


mariobehling:
Thank you!

PR #676, opened 2017-08-05
Fixes #585 - Improvise websearch rss results

### LINKED ISSUE #585: Increase the size of web search box
Please improve the results as follows
* make the search result function more similar to Google Assistant. 
* Also increase height and add images if possible.
* Use the full width for the search result output.
* Add the info about the URL of the website

uday96:
Fixes issue #585 

**Changes:**
 - [x] Preview the link to fetch images for rss
 - [x] Use full width to show results
 - [x] Make result tiles UI similar to google
 - [x] Add the info about the URL of the website

**Demo Link:** http://styleswipe.surge.sh/

**Screenshot**

swip


uday96:
@rishiraj824 @isuruAb @madhavrathi Please review.

We are not getting images from server, so I am using the endpoint `BASE_URL+'/susi/linkPreview.json?url=URL'` for fetching images for each url and it is resulting in too many ajax calls and also the server is throwing a `503 Request frequency too high error` for many requests. I think the only solution to this is to preview the links on server side only and the client directly uses that to render cards.

Any suggestions?

rishiraj824:
@uday96 I totally agree, too many ajax calls already in the application. Instead, a common endpoint for the image will be better.

uday96:
Through https://github.com/fossasia/susi_server/pull/506 , some of the answers[0].data objects have `image` in the server response. We now first use the data elements which already have image and for the remaining we use the linkPreview service - `http://api.susi.ai/susi/linkPreview.json?url=URL` to get the images. This reduces the total ajax calls being made

uday96:
@rishiraj824 @isuruAb @madhavrathi @mariobehling Please review

uday96:
@isuruAb @madhavrath Made the changes
 
a

Thanks.

mariobehling:
Ok, let's test this in everyday settings. Thanks!

PR #670, opened 2017-08-03
Fixes: #664 - Store Custom theme settings in server

isuruAb:
Fixes issue #664 

Changes: 

- [x] Store Custom theme colours in server
- [x]  Load default theme data 
- [x] load theme data from server 

Demo Link: http://isuruab.surge.sh/

Screenshots for the change:

PR #689, opened 2017-08-06
Fixes #573 - Use a single server call to update all settings on server

### LINKED ISSUE #573: Use new settings endpoint to send multiple settings to server in one call
**Actual Behaviour**

Currently we are making a sever call for every setting changed

**Expected Behaviour**

Use https://github.com/fossasia/susi_server/pull/404 and send all settings at once to server reducing the number of server calls

Yes

uday96:
Fixes issue #573 

**Changes:**
 - We now make a single server call to update all our settings on server instead of a server call for each setting

**Demo Link:**  http://susisettings.surge.sh/

Settings being used : 
```
                        theme: string
			enterAsSend: boolean
			micInput: boolean
			speechOutput: boolean
			speechOutputAlways: boolean
			speechRate: 0.5-2 (decimal)
			speechPitch: 0-2 (decimal)
                        hotwordRecognition: boolean (required for other clients)
                        ttsLanguage: string
                        prefLanguage: string
                        customThemeValue: string
```

@rishiraj824 @isuruAb @madhavrathi @saurabhjn76 Please review


src/actions/API.actions.js
```diff
@@ -407,12 +338,14 @@ export function pushCustomThemeToServer(customTheme){
     return;
   }
        url = BASE_URL+'/aaa/changeUserSettings.json?'
-          +'key=custom_theme_value&value='+customTheme
-          +'&access_token='+cookies.get('loggedIn');
+          +'key1=customThemeValue&value1='+customTheme
```
rishiraj824: why are we sending customThemeSettings separately?

uday96: It was implemented in https://github.com/fossasia/chat.susi.ai/pull/670 by @isuruAb . The custom theme call is made to server from the theme dialog and not from settings route. So sending it seperately

rishiraj824: @uday96 Okay thanks. I was just confirming.

hpdang:
Thanks.

PR #743, opened 2017-08-19
Fixes #742 - Add Page "Privacy" and take out "Settings" from Footer

### LINKED ISSUE #742: Add Page "Privacy" and take out "Settings" from Footer
We need a page for a privacy policy. Use the same style as for "Terms" page. Also do the following:
* Add link "Privacy" in the footer on the right hand side
* Implement the following order: Privacy, Terms, Contact
* Take out "Settings" from footer

screenshot from 2017-08-19 16-33-48

rishiraj824:
Fixes issue #742 

Changes: 
- [x] Added a page for a privacy policy. 
- [x] Used the same style as for "Terms" page.
- [x] Added link "Privacy" in the footer on the right-hand side
- [x] Implemented the order: Privacy, Terms, Contact
- [x] Took out "Settings" from footer

- [ ] Content for Privacy Policy needs to be added. Currently, it's the Terms page

Demo Link

http://susi-privacy.surge.sh/privacy

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -12,6 +12,7 @@ import Settings from './components/ChatApp/Settings/Settings.react';
 import Support from './components/Support/Support.react';
 import Team from './components/Team/Team.react';
 import Terms from './components/Terms/Terms.react';
+import Privacy from './components/Privacy/Privacy.react';
 import { Switch, Route } from 'react-router-dom';
 import React, { Component } from 'react';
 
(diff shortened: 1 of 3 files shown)


DravitLochan:
@rishiraj824 Shouldn't this information come from server?

DravitLochan:
@madhavrathi I checked with other pages. None of them has data that needs to be shown on other clients. Still, if we come across any such page or piece of data, that still needs to be coming from the server.

madhavrathi:
@DravitLochan that is true. we need this on all clients.

mariobehling:
Thanks!

Issue #819, opened 2017-09-22
Settings: Improve Save button actions


hpdang:
On the settings tabs please change the save button action as follows:
* Show save button in active stage only when something can be saved
* Show save button inactive when there is nothing to save
* Stay on the same tab after user presses "Save"

![screenshot from 2017-09-22 11-07-14](https://user-images.githubusercontent.com/1658742/30737393-b0a5838a-9f86-11e7-8ccb-b07562b9c209.png)

hkedia321:
I would like to work on it :+1:

hkedia321:
In the settings page, the **"SAVE CHANGES" button looks redundant** for the following tabs:
1. Server Settings (when logged in)
2. Account Settings (when logged in)
3. Connect to SUSI Hardware 

This is because we have to edit the settings in a popup having its own save button. So the Save button on the settings page has no use.
**Can we remove these buttons on these tabs ?**
Please provide me suggestions on this.
Thanks.

zamhaq:
@mariobehling @rishiraj824 this issue can be closed?

rishiraj824:
closing via #836

PR #836, opened 2017-09-26
Fixes issue #819 - Improve save button actions in settings

hkedia321:
Fixes issue #819 

Changes:
- Show save button in active stage only when something can be saved
- Show save button inactive when there is nothing to save
- Stay on the same tab after user presses "Save"
- Hide save button from the following tabs, since they are redundant : 
    - Server Settings (when logged in)
    - Account Settings (when logged in)
    - Connect to SUSI Hardware
- Load Initial settings when user changes the tab
- select the theme option based on the actual value rather than having an initial value
- Set the document title in settings page

**Demo Link:** [sneaky-education.surge.sh](http://sneaky-education.surge.sh/)

Screenshots for the change: 
image

### DIFF
--- a/src/components/ChatApp/Settings/Settings.react.js
+++ b/src/components/ChatApp/Settings/Settings.react.js
@@ -39,8 +39,8 @@ const cookies = new Cookies();
 
 class Settings extends Component {
 
-	constructor(props) {
-		super(props);
+	// save a variable in state holding the initial state of the settings
+	setInitialSettings = () => {
 		let defaults = UserPreferencesStore.getPreferences();
 		let defaultServer = defaults.Server;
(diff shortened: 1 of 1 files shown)


hkedia321:
@madhavrathi @rishiraj824 Please review this PR. Thanks.

mariobehling:
Actually, it might be out of scope. So, I am merging this. Cheers! Good work.

mariobehling:
Please follow up here, if you like: https://github.com/fossasia/chat.susi.ai/issues/840

hkedia321:
@mariobehling, Sure I will work on #840 and further improve the settings page. :+1:

Issue #823, opened 2017-09-23
Bottom text field disappears when search box on top is activated


mariobehling:
The bottom text field disappears when search box on top is activated. 

Expected: Bottom field should always stay.

![screenshot from 2017-09-23 16-07-48](https://user-images.githubusercontent.com/1583873/30773850-8500b788-a079-11e7-967a-1998d42b0ea3.png)

RMCoder198:
@mariobehling  I would like to work on this issue.
  Will you guide me how can i find the file from which i can fix this issue.
I am new to open source contribution.

madhavrathi:
@RMCoder198 https://github.com/fossasia/chat.susi.ai/blob/master/src/components/ChatApp/SearchField.react.js This is the file you should look for. This was initially intended as a feature. you can disable the hiding of text box in this file. Cheers!

RMCoder198:
@madhavrathi  Thanks .

PR #842, opened 2017-09-30
message composer will mount onClick search

nileshgulia1:
fix for issue #823


nileshgulia1:
The issue also fixes when i use``` 
 {this.componentDidMount}``` instead of ```{this.onStart}```

Issue #838, opened 2017-09-28
Avoid unnecessary API calls


mariobehling:
Loading the web UI shows that the Google Translate API gets called. Please avoid an integration with other services and take out the API call.

Also avoid calling Google web fonts. Instead add fonts locally.

Check out if any other external API calls are made for no reason and remove them.

zamhaq:
@mariobehling I would like to work on this.

zamhaq:
@madhavrathi @mariobehling can anyone help with deployment to surge? I tried deploying at hallowed-branch.surge.sh but this gives 404.

madhavrathi:
@zamhaq set your upload path to build folder. That should do the thing. Pls send screenshot if error still persists.

isuruAb:
@zamhaq  can we close this issue ?

zamhaq:
@isuruAb I removed the calls the `gtranslate` but I could not find the source calling `web fonts`.

PR #845, opened 2017-10-02
Remove unnecessary api calls to gtranslate

zamhaq:
Fixes issue #838 

Changes: Removed multiple calls to google translate API on load of chat app.

Demo Link: http://neat-sack.surge.sh/ (updated)

Screenshots for the change: 
As the network tab shows, gtranslate only gets called when sending or receiving messages.

gtranslate

### DIFF
--- a/src/actions/API.actions.js
+++ b/src/actions/API.actions.js
@@ -118,26 +118,6 @@ export function createSUSIMessage(createdMessage, currentThreadID, voice) {
         // Setting Language received from User
         receivedMessage.lang = response.answers[0].actions[0].language;
       }
-      let defaultPrefLanguage = defaults.PrefLanguage;
-      // Translate the message text
-        let urlForTranslate = 'https://translate.googleapis.com/translate_a/single?client=gtx&sl=en-US&tl='+defaultPrefLanguage+'&dt=t&q='+receivedMessage.text;
-        $.ajax({
-          url: urlForTranslate,
-          dataType: 'json',
-          crossDomain: true,
-          timeout: 3000,
-          async: false,
-          success: function (data) {
-            if(data[0]){
-              if(data[0][0]){
-                receivedMessage.text = data[0][0][0];
-              }
-            }
-          },
-          error: function(errorThrown){
-            console.log(errorThrown);
-          }
-        });
       receivedMessage.response = response;
       let actions = [];
       response.answers[0].actions.forEach((actionobj) => {
--- a/src/actions/History.actions.js
+++ b/src/actions/History.actions.js
@@ -80,52 +80,11 @@ export function getHistory() {
         userMsg.date = new Date(cognition.query_date);
         userMsg.text = query;
 
-        // Translate history
-        let defaultPrefLanguage = defaults.PrefLanguage;
-        // Translate the User message text
-        let urlForTranslate = 'https://translate.googleapis.com/translate_a/single?client=gtx&sl=en-US&tl='+defaultPrefLanguage+'&dt=t&q='+userMsg.text;
-        $.ajax({
-          url: urlForTranslate,
-          dataType: 'json',
-          crossDomain: true,
-          timeout: 3000,
-          async: false,
-          success: function (response) {
-            if(response[0]){
-              if(response[0][0]){
-                userMsg.text = response[0][0][0];
-              }
-            }
-          },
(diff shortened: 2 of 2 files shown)


zamhaq:
@mariobehling @madhavrathi please review.

zamhaq:
@rishiraj824 so the multiple calls to `gtranslate` on the initial load of the webchat is for speech demonstration 
```
export function getTTSLangText(voiceList){
  let defText = 'This is an example of speech synthesis';
  voiceList.forEach((voice,index) => {
    let url = 'https://translate.googleapis.com/translate_a/single?client=gtx&sl=en-US&tl='+voice.lang+'&dt=t&q='+defText;
    $.ajax({
      url: url,
      dataType: 'json',
      crossDomain: true,
      timeout: 3000,
      async: false,
      success: function (response) {
        if(response[0]){
          if(response[0][0]){
            let translatedText = response[0][0][0];
            voice.translatedText = translatedText;
          }
        }
      },
      error: function(errorThrown){
        console.log(errorThrown);
      }
    });
  });
}
```
But adding these lines causes the follwing example calls to gtranslate api on initial load as seen from the network tab :- 
multgtapi

The lines you suggested for change are the ones that translate the text message when a message is received according to the `default pref language`. I don't think they are the ones that cause the multiple example calls if I am not wrong.

Should I try to fix the API calls by making a request with example text only when speech demonstration example is requested?

What should I do? Please rectify me if I am wrong @rishiraj824 @madhavrathi @mariobehling.

tabesin:
Architecture wise I would assume that you only use 1 to 2 different languages. So the bulk of translations is irrelevant.

Also the line

> let defText = 'This is an example of speech synthesis';

sounds like they copy pasted without an orphaned functionality.

rishiraj824:
@tabesin @zamhaq The code you removed is to test the 'Speech output in the given languages'. It's not the one which comes up while the site loads. The history which loads also sends a request to the Google Translate API which converts the messages into German if I have German selected in the Text Language Setting. https://github.com/fossasia/chat.susi.ai/blob/master/src/actions/History.actions.js#L85 Check this line as well.

zamhaq:
@rishiraj824 I understand that the history which loads also sends api calls to google translate. 

But, on the initial load if you have the lines which i removed then you can see on the networks tab having requests like 'This is an example of speech synthesis' to google translate. Could you please check with the current master version and check the networks tab to see those requests @tabesin @rishiraj824

rishiraj824:
@zamhaq Those lines are not at all unnecessary and they ought to be there. Please try to understand this. It is required to change the Speech Output settings. The unnecessary Google Translate calls are the ones which I have pointed out. Please only remove them. Thank you.

zamhaq:
@rishiraj824 ok. So I need to remove the lines https://github.com/fossasia/chat.susi.ai/blob/master/src/actions/API.actions.js#L122 and https://github.com/fossasia/chat.susi.ai/blob/master/src/actions/History.actions.js#L85? Thanks.

zamhaq:
@rishiraj824 Please review now. 

Removed calls to gtranslate from `History.actions.js`(https://github.com/fossasia/chat.susi.ai/blob/master/src/actions/History.actions.js#L85) which resolves the calls to gtranslate for already sent texts.

Removing https://github.com/fossasia/chat.susi.ai/blob/master/src/actions/API.actions.js#L122 lines result in Susi's response not being translated.

rishiraj824:
@zamhaq Thank You. Please put up a demo link for the reviewers to test it.

zamhaq:
@rishiraj824 updated the demo link above.

rishiraj824:
@zamhaq Please make the changes to both the files API.actions.js as well.

zamhaq:
@rishiraj824 remove these lines https://github.com/fossasia/chat.susi.ai/blob/master/src/actions/API.actions.js#L122?

rishiraj824:
@zamhaq Yes, those should be removed too. I think right now translating it using Google Translate is not the right way. We need to find a way to this on the server end.

zamhaq:
@rishiraj824 Ok I will remove those too.

zamhaq:
@rishiraj824 I have updated the demo link after removing https://github.com/fossasia/chat.susi.ai/blob/master/src/actions/API.actions.js#L122. Please review again for any problems. Thanks.

PR #851, opened 2017-10-05
Fixes #835 : Theme changes doesn't change Mic icon

akshatnitd:
Fixes issue #835 

Changes: 
* Make correction in the state name that was responsible for setting the theme name
* Send correct key name to the request to the Server. It was incorrectly sent as **Theme**, instead of **theme**.

[Demo Link](http://cooperative-twist.surge.sh)

Screenshots for the change: 
sscreen

PR #853, opened 2017-10-05
Fixes #852 : Background color changes with theme change

### LINKED ISSUE #852: Background color resets to White everytime the theme is changed
**Actual Behaviour**

The body color changes to white after making the changes from the *Change Theme* option on the top-right drop down menu.

**Expected Behaviour**

It should be of the color, what the user has selected from the color picker

**Steps to reproduce it**
* Login to the platform
* Change the body color from *Change Theme* option from the top-right Drop Down Menu
* Click Submit 
You can see the page background color changes to White
http://cooperative-twist.surge.sh/

**Screenshots of the issue**
sscreen

Yes, I would lke to work on it

akshatnitd:
Fixes issue #852 

Changes:
* There was a mistake where the background color was assigned. The background color was assigned before setting the value of the variable.
* The order of the code was corrected inside the render function.

Screenshots for the change: 
screenshot

[Demo link](http://cooperative-twist.surge.sh)

### DIFF
--- a/src/components/ChatApp/MessageSection/MessageSection.react.js
+++ b/src/components/ChatApp/MessageSection/MessageSection.react.js
@@ -541,36 +541,34 @@ class MessageSection extends Component {
     var messagePane;
     var textArea;
     var buttonColor;
+
+    switch(this.state.currTheme){
+      case 'custom':{
+        bodyColor = this.state.body;
+        TopBarColor = this.state.header;
+        composerColor = this.state.composer;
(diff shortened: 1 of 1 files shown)


akshatnitd:
@isuruAb Demo link added

akshatnitd:
@isuruAb Made relevant changes to #851. This can be merged after #853

isuruAb:
@mariobehling  please check this

PR #858, opened 2017-10-14
Fix #818 - Copy previous messages on pressing 'up' and 'down' arrow keys

### LINKED ISSUE #818: Copy previous message in chat on pressing the "up" key
**Actual Behaviour**

Nothing happens on pressing the "up" or the "down" key in chat

**Expected Behaviour**

On pressing the "up" key the previous message should be copied. And on pressing the "down" key, the following message should be copied.
Often the user wants to copy the previous message. This would enable the user to easily send previous messages with little or no modification to the message. 

**Steps to reproduce it**

Open the chat, and press the "up" or "down". Nothing happens.

Yes, I would like to work on this feature.

hkedia321:
Fixes issue #818 

Changes:
- Copy the user's previous message on pressing 'up' key
- Copy the user's next message on pressing 'down' key

Demo :[overjoyed-skin.surge.sh](http://overjoyed-skin.surge.sh/)

Screenshots for the change: 
image

### DIFF
--- a/src/components/ChatApp/MessageComposer.react.js
+++ b/src/components/ChatApp/MessageComposer.react.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import Send from 'material-ui/svg-icons/content/send';
 import Mic from 'material-ui/svg-icons/av/mic';
 import UserPreferencesStore from '../../stores/UserPreferencesStore';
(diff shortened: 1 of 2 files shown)


mariobehling:
Nice. Thank you.

Issue #863, opened 2017-10-16
onspeechend is never called


zamhaq:
**Actual Behaviour**

When speech ends, `onspeechend` is not called.

**Expected Behaviour**

When speech ends, `onspeechend` should be called.

**Steps to reproduce it**

If we check the console for events being fired, then only the `onEnd` function is called when `end` event fires. But, `onspeechend` is never called. Test by adding `console logs` to both the listeners.

**Screenshots of the issue**

![onendcode](https://user-images.githubusercontent.com/20766601/31628379-22680efc-b2ce-11e7-8427-58c2b2db4050.png)

![onend](https://user-images.githubusercontent.com/20766601/31628357-1902924c-b2ce-11e7-9653-5eea346f068a.png)

**Would you like to work on the issue?**

Yes.

PR #867, opened 2017-10-20
Remove listener for onspeechend

zamhaq:
Fixes issue #863 (onspeechend is never called)

Changes: Remove event listener `onspeechend` function which was listening to the wrong event `onspeechend` instead of `speechend`. Listening to the correct event `speechend` would result in two `speech to text` results generated since `onspeechend` was calling `onEnd`, but that is not required in this case since `onEnd` is automatically invoked when `speech recognition ends`.

Demo Link: https://silent-language.surge.sh/

### DIFF
--- a/src/components/ChatApp/MessageComposer.react.js
+++ b/src/components/ChatApp/MessageComposer.react.js
@@ -74,11 +74,8 @@ class MessageComposer extends Component {
     })
   }
 
-  onspeechend = () => {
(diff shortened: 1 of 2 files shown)


zamhaq:
Closing this as I have opened a new PR https://github.com/fossasia/chat.susi.ai/pull/870. Please refer there.

PR #880, opened 2017-11-01
Fix #879 - set previous theme on exit

### LINKED ISSUE #879: Custom theme colours do not change to default colours when user clicks the close button
**Actual Behaviour**

Custom theme colours do not change to default colours when user clicks the close button

**Expected Behaviour**

It should change to previously selected colour theme when user clicks on the close button (when user does not need to save selected theme information)

**Steps to reproduce it**

login > select custom theme colours > click close button of custom theme colour dialog box

anyone can get it

hkedia321:
Fixes issue #879

**Changes:**
- have a state variable `prevThemeSettings` to store the previous theme on opening of the dialog
- apply the theme from `prevThemeSettings` on closing the dialog
- avoid applying the above step when user "saves" a new theme

**Result:** On closing the change theme dialog, the previously saved settings in applied instead of the unsaved changes.

**Demo Link:** [saveprevtheme-susi.surge.sh](http://saveprevtheme-susi.surge.sh)

@isuruAb @rishiraj824 @madhavrathi Please review.

### DIFF
--- a/src/components/ChatApp/MessageSection/DialogSection.js
+++ b/src/components/ChatApp/MessageSection/DialogSection.js
@@ -101,6 +101,7 @@ DialogSection.propTypes = {
     actions: PropTypes.object,
     bodyStyle: PropTypes.object,
     onRequestClose: PropTypes.func,
+    onSaveThemeSettings: PropTypes.func,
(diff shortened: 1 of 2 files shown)


mariobehling:
Thanks.

Issue #948, opened 2017-12-08
Move images in public folder into "images" in src folder


mariobehling:
All source files including images should be in the src folder.

Remorax:
I would like to take this up :)

marvin08:
@mariobehling I will do it. :+1:

marvin08:
@Remorax May I create a PR as I am done with it?

marvin08:
@Remorax May I create a PR as I am done with it?

Remorax:
Oh, I didn't notice your comment. I was in the process of creating a PR myself :P

marvin08:
@Remorax :+1:

anshumanv:
@Remorax will you be sending another PR or can I send one? :smile:

Remorax:
Sure, go ahead @anshumanv :)

anshumanv:
Alright, thanks @Remorax :+1:

anshumanv:
@mariobehling the scaffolding tool we use is create-react-app which has some restrictions attached. They expect all of the static files(which are copied directly while bundling the app) like favicon etc in the public folder and once we run `npm run build` to package a production build these files are directly copied to the `build` folder that is generated. So moving favicon to `src/images` isn't possible with the current version of CRA. For reference please see - https://github.com/facebookincubator/create-react-app/issues/585

abishekvashok:
@anushmanv we could have pre compile script that we should run before npm install in the appjson which copies images to public folder for us and at last cleans up the folder, what do you think?

anshumanv:
We can always go for such workarounds, but I'd suggest sticking to the one suggested by cra.

abishekvashok:
> cra.

???

anshumanv:
Create React App.

abishekvashok:
@mariobehling status on this

0x48piraj:
I can work on this, the other way. **(Workaround)**

PR #949, opened 2017-12-08
Removed images in public folder

Remorax:
Fixes issue #948 

Changes: 
The images in public folder have been deleted as duplicates exist in src/images.


mariobehling:
Why are you deleting the images. They are supposed to be moved and the code needs to be updated to the new location. Please submit a new PR, that solves the issue. Closing it therefore.

Issue #984, opened 2017-12-30
Add Russian as an option in Text Language Settings


arnav-mandal1234:
### Actual Behaviour

Missing option for Russian language.

### Expected Behaviour

Add Russian as an option.

### Steps to reproduce it

Go to settings then Text language settings.

### Would you like to work on the issue?

Yes

0x48piraj:
Close this @arnav-mandal1234 :+1:

PR #985, opened 2017-12-30
Added Russian as an option in Text Language settings

arnav-mandal1234:
Fixes issue #984

Changes: Added Russian as an option in Text Language Settings


arnav-mandal1234:
@isuruAb @anshumanv @Remorax @aashutoshrathi check into this.

Issue #986, opened 2017-12-31
Auto deploy to surge after travis passes


isuruAb:
**Actual Behaviour**

Now PR owner has to manually do the surge deployment if he/she need to provide a preview link

**Expected Behaviour**

We can automate this and  deploy automatically on surge after tests are passed.

**Would you like to work on the issue?**

yes

ParthS007:
@isuruAb Please close this Issue. 👍

PR #989, opened 2018-01-01
Deploy every PR on surge

isuruAb:
Fixes issue #986
Changes: add surge deployment shell file
Demo link: https://pr-989-fossasia-susi-web-chat.surge.sh/


ParthS007:
@isuruAb  LGTM!
Just a suggestion, if you update the PR template such that, after which contributor have to write only PR number will be more useful as at present PR link is generated in Travis log and every time we have to open it for review to check the changes. Just like this - https://github.com/fossasia/susper.com/pull/852

isuruAb:
@ParthS007  yes it is good suggestion . I did those changes please check.

ParthS007:
Thanks @isuruAb  😄  for considering it 👍

Issue #996, opened 2018-01-02
Bugs in Share to Facebook section


Remorax:
**Actual Behaviour**

There are a few bugs in the Share to Facebook section, such as:
1. The share button has black label with white background, whereas the colour combination used throughout is white label with blue background
2. The share method is called using onClick, instead of onTouchTap as it has been used throughout this site.
3. The Share button labels and the title do not get translated.
4. keyboardFocused has not been set to true

**Expected Behaviour**

These bugs need to be fixed.

**Steps to reproduce it**

These bugs can be seen by going to the Share on Facebook section in Settings.

**Screenshots of the issue**

![screen shot 2018-01-02 at 8 09 23 pm](https://user-images.githubusercontent.com/26062692/34487121-ed2d04e6-eff8-11e7-99d3-7c23c9f7b22d.png)

**Would you like to work on the issue?**

Yes, I would like to work on this.

ParthS007:
@Remorax Please close this Issue 👍

PR #997, opened 2018-01-02
Fixed the bugs in the share on Facebook section

Remorax:
Fixes issue #996 

Changes:
1. The background color and the label color of the share button was changed.
2. The fields for "Share" and "Share about SUSI on Facebook" were added to the JSON files of all languages.
3. onClick was changed to onTouchTap and keyboardFocused was set to True.

Demo Link: http://succinct-exchange.surge.sh

Screenshots for the change: 
screen shot 2018-01-02 at 9 17 57 pm

@madhavrathi @isuruAb @rishiraj824 Please review ASAP 😄


mariobehling:
Ok, rather than having a share on Facebook thing, we should rather have an apps integration with SUSI. But, let's see this as work in progress. Cheers!

Issue #1004, opened 2018-01-06
Reset button after implementing custom theme reverts to light theme


Remorax:
**Actual Behaviour**

The reset button is the themes section reverts the theme back to the light theme, regardless of whether the theme before making the changes, was dark or light.

**Expected Behaviour**

The reset button should reset the theme to what it was before the changes were made, and not to the light theme always.

**Steps to reproduce it**

1. Go to Themes, and make sure that the current theme is dark (otherwise go to Settings and change it to dark).
2. Change, say the header colour, or anything for that matter such that the theme is custom.
3. Click on Reset.
4. It reverts to light theme, instead of the dark theme like it should.

**Screenshots of the issue**

![screen shot 2018-01-07 at 12 50 37 am](https://user-images.githubusercontent.com/26062692/34643272-983da164-f347-11e7-8636-cc718402845d.png)
![screen shot 2018-01-07 at 12 50 50 am](https://user-images.githubusercontent.com/26062692/34643273-9867e604-f347-11e7-9a8b-7a4870535d89.png)
![screen shot 2018-01-07 at 12 51 01 am](https://user-images.githubusercontent.com/26062692/34643274-98987b66-f347-11e7-9185-98ef160bed59.png)

**Would you like to work on the issue?**

I would like to work on this.

PR #1005, opened 2018-01-06
Fixed the reset button bug

Remorax:
Fixes issue #1004 

Changes:
On clicking the reset button, the theme is reverted to what it was before the changes were made.

Demo Link: https://pr-1005-fossasia-susi-web-chat.surge.sh


anshumanv:
It still seems to revert to light theme, perhaps isn't reset button meant to discard all changes in the theme and revert it back to what we started with i.e Light theme?

Remorax:
> It still seems to revert to light theme

@anshumanv That is strange, I rechecked and it works perfectly fine for me.

>  perhaps isn't reset button meant to discard all changes in the theme and revert it back to what we started with i.e Light theme?

That's the thing, what if the user didn't start out with the light theme? What if the user prefers the dark theme, in that case, considering the light theme as the original would be wrong, no?
I for instance always use the dark theme, but when I make changes to the custom theme and press Reset, I expect it to go back to what it was before I made the changes, that is, dark theme. That's what Reset means by definition, no?

anshumanv:
> @anshumanv That is strange, I rechecked and it works perfectly fine for me.

Not sure then, please let others check in that case. :smile:

Remorax:
@aashutoshrathi I'll rebase it no problems. But I'm more concerned as to why you're not able to view the changes. It works perfectly fine for me, I've checked with both the accounts I have. After making the changes, when Reset button is pressed, then only the theme reverts back to the original. That is, if the theme was light originally, it reverts back to Light otherwise it reverts back to Dark. Can you check if this works for you? I'll rebase after that. Thanks 😄

madhavrathi:
@Remorax Please check travis

saurabhjn76:
@Remorax Please check the Travis

Remorax:
@saurabhjn76 @madhavrathi It says that I do not have permission to deploy to the URL and that's why its' failing in 1 of 3 checks in Travis. However I have already deployed it to this very same URL: https://pr-1005-fossasia-susi-web-chat.surge.sh. So I think this PR can be merged, if its' fine otherwise. Thanks 😄

Remorax:
@isuruAb @madhavrathi @mariobehling @saurabhjn76 Can this be merged then? Its been 23 days since I made this PR 😅

madhavrathi:
@Remorax Please check travis errors.

Remorax:
Its failing because it gives an error saying I do not have permission to deploy to the URL, and I think that can be ignored since I've already deployed it.
So can it be merged?@madhavrathi @rishiraj824 @saurabhjn76 
screen shot 2018-01-31 at 3 43 04 pm

src/components/ChatApp/MessageSection/MessageSection.react.js
```diff
@@ -40,8 +40,8 @@ function getStateFromStores() {
     SnackbarOpenBackground: false,
     messages: MessageStore.getAllForCurrentThread(),
     thread: ThreadStore.getCurrent(),
-		currTheme: UserPreferencesStore.getTheme(),
-		tour:true,
```
daminisatya: Next time make sure you don't commit the indentation changes made by your iDE. Make sure you commit only your changes.

PR #1024, opened 2018-01-10
Fixes #1023. Set a default response message for susi when cognition fails.

### LINKED ISSUE #1023: Infinitely loading gif when susi is not able to answer a query.
**Actual Behaviour**

image

**Expected Behaviour**
Susi should respond with "I don't understand" instead of displaying an infinitely loading gif.

**Steps to reproduce it**
Send gibberish text or a very long text to susi at [https://chat.susi.ai/](https://chat.susi.ai/)

Yes

kapillamba4:
Fixes issue #1023.

Changes: Set a default response message for susi instead of displaying an infinitely loading gif as a chat message by modifying the ajax call to susi server.

Demo Link: https://pr-1024-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 
image

### DIFF
--- a/src/actions/API.actions.js
+++ b/src/actions/API.actions.js
@@ -43,7 +43,24 @@ export function getLocation(){
 }
 // Main server call for Creating a SUSI Message
 export function createSUSIMessage(createdMessage, currentThreadID, voice) {
-  var timestamp = Date.now();
+  const timestamp = Date.now();
+  const timezoneOffset = (new Date()).getTimezoneOffset();
+  const defaultAnswer = {
+    data: [{
+      '0': '',
+      timezoneOffset,
+      language: 'en'
+    }],
+    metadata: {
+      count: 1
+    },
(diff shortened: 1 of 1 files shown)


kapillamba4:
@aashutoshrathi I am looking into the issue, please wait.

kapillamba4:
@akshatnitd and @aashutoshrathi  Chat persistence is not possible in case of cognition failure, the problem lies at the server side as the chat messages are stored as a json file there and it is automatically updated with each message sent and recieved from this chat app.With or without this patch our message is still not saved at the server if the cognition fails as you can see below:
Before:
image
After:
image

But i think this is still better than showing an infinitely loading gif.
Our message is lost either way, with or without this patch.

kapillamba4:
updated the deployment link 👍

kapillamba4:
@madhavrathi, can you review and merge this please 👍

PR #1096, opened 2018-02-07
Solve toggle switch issue in settings

### LINKED ISSUE #1070: Settings are changing by clicking on text along with toggle switch.
**Actual Behaviour**

In SUSI settings, we're able to change settings even if we click on text on the left of the toggle switch.

**Expected Behaviour**

There's a toggle switch to change settings so it should change only when we click on toggle switch.

**Steps to reproduce it**

Go to https://chat.susi.ai/settings and try to change settings under options "ChatApp Settings", "Mic Settings" or "Speech Settings". This problem is with all of them. You'll see that you're able to change settings even if you click on the text and not on the toggle switch. This shouldn't happen.

**Screenshots of the issue**

This issue exists for -
Under "ChatApp Settings"
screen shot 2018-01-26 at 5 15 42 pm

Under "Mic Settings"
screen shot 2018-01-26 at 5 15 50 pm

Under "Speech Settings"
screen shot 2018-01-26 at 5 15 58 pm

It's correct for "Server Settings" -
screen shot 2018-01-26 at 5 16 08 pm

Yes

DNS-404:
Fixes #1070

Changes: Now switch will be toggled only when we click on the switch and not when we click on the text.(Please refer issue)

Demo Link: http://burly-hole.surge.sh/
Open this link and go to settings.
The toggle switches in menus "ChatApp Settings", "Mic Settings" and "Speech settings" are now triggered only when we click on the switch and not when we click on the text with it.

### DIFF
--- a/src/components/ChatApp/Settings/Settings.css
+++ b/src/components/ChatApp/Settings/Settings.css
@@ -36,13 +36,13 @@
 }
 
 .settingsBtns {
-  top: 10px;
-  margin-left: 30px ;
+  top: 5px;
+  margin-left: 20px ;
 }
 
 .settings-toggle {
-  padding: 20px;
-  max-width: 60%;
+  max-width: 5%;
+  float: left;
 }
 
 .settingsSubmit{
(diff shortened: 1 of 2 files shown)


DNS-404:
Please review this.

PR #1106, opened 2018-02-24
Handles angle brackets

### LINKED ISSUE #1042: [BUG] Susi doesn't take input of strings starting with angle bracket, random results
**Actual Behaviour**
Susi doesn't take input of strings starting with `<` and **responds randomly to it** 

**Expected Behaviour**
Susi App must either reject the input before sending it to the `susi_server` , or
display the message after escaping it instead of **completely not showing it**.
An error message can also be added for such input.

**Steps to reproduce it**
- Add checks for input
- Display notification for improper input format

**Screenshots of the issue**
susi-bug

Not too comfortable in react. It should be assigned to someone else.

sayak119:
Fixes #1042 

Changes: Handles angle brackets properly and doesn't show random results. Also added regex to handle other rogue cases which were found during further testing of the bug.

Demo Link: http://debonair-nail.surge.sh/

Screenshots for the change: 

screencast_sunday-25-february-2018_02_50_16-ist


sayak119:
@hkedia321 @anshumanv please review

sayak119:
@hkedia321 when we use `<b>hello</b>` then we get stuck with an infinite gif. My question is that do we even need this bold feature ?

screenshot from 2018-02-27 04-35-10

screenshot from 2018-02-27 04-37-57

screenshot from 2018-02-27 04-37-42

sayak119:
@isuruAb @anshumanv

sayak119:
@hkedia321 @anshumanv my solution now handles all of the cases.

sayak119:
@hkedia321 yup as of now it works but also the HTML tags are also not parsed properly as shown in the screenshots. They either go into infinite gifs or return some random answers which sometimes have no connection with the query.

sayak119:
@rishiraj824 @madhavrathi @isuruAb please review

sayak119:
@Orbiter it will perform the mathematical operations well. I have kept a proper check for that. As for spaces, I have started working on it.

sayak119:
@Orbiter please review

Orbiter:
does 
```
else {
    messageCheck = message.text.replace(/^[`~><%^&!@#*()_|+=;:"'<>\{\}\[\]\\\/]*/, ' ');
    messageCheck = messageCheck.replace(/[`~><$%^&@#*(_|+=:;<>"'\{\}\[\]\\\/]*$/, ' ');
  }
```
mean that all those characters are replaced? That also means no mathematical expression is possible. 
This issue was opened just about angle brackets, now the process removes everything non-ascii?

The server handles at least the following characters, which are removed here:
+=_()!%^@#&|
because of special uses or part of mathematical expressions.

Why is there such a complex expression about html entities? Angle brackets may cause a security issue for XSS attacks and must be handled therefore. But that applies only for JavaScript injection, so that should be checked.

stealthanthrax:
@sayak119 , would you like to follow up here?

mariobehling:
Closing this due to inactivity.

PR #1132, opened 2018-03-05
Live Theme preview 

### LINKED ISSUE #1107: No preview when changing theme.
**Actual Behaviour**
When the user selects a theme in the settings, no preview is provided unless the "Save Changes" is pressed, which requires same steps to revet back to the original theme, if the user does not like the new one. 

**Expected Behaviour**

The theme should immediately change when ever the corresponding radio button is selected, and reverted to original if the user navigates away without pressing the "Save Changes" button.

**Steps to reproduce it**

Select a theme in the settings menu.

Yes

mabrar21:
Fixes #1107 

Changes: Added functionality for theme preview as soon as the user selects a 
theme, and reverts back to last saved theme if the setting are not saved.
Implemented by saving the state of current theme into original Theme variable, 
and changing it only when user saves settings.

Demo Link: https://theme-preview.surge.sh

Screenshots for the change: 
screenshot from 2018-03-05 06-31-38
screenshot from 2018-03-05 06-31-15

### DIFF
--- a/src/components/ChatApp/Settings/Settings.react.js
+++ b/src/components/ChatApp/Settings/Settings.react.js
@@ -44,13 +44,14 @@ import ShareIcon from 'material-ui/svg-icons/social/share';
 const cookies = new Cookies();
 
 class Settings extends Component {
-
+	// Boolean to store the state of preview i.e which theme to display
+	preview = false;
 	// save a variable in state holding the initial state of the settings
 	setInitialSettings = () => {
 		let defaults = UserPreferencesStore.getPreferences();
 		let identity = UserIdentityStore.getIdentity();
 		let defaultServer = defaults.Server;
-		let defaultTheme = defaults.Theme;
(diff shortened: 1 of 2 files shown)


mabrar21:
Thank you!

mabrar21:
@anshumanv Once you save a theme, the save button is supposed to be disabled, since there are no  further "unsaved settings" to be saved. Can you explain how did you create the persisting theme scenario? Whenever I switch to another settings' tab, the theme is reverted to original one.

anshumanv:
> Once you save a theme, the save button is supposed to be disabled, since there are no further "unsaved settings" to be saved.

I know this much m8 :smile: 
peek 2018-03-07 20-08

Clicking on other theme tab reverts the theme, try clicking on the susi icon, please handle this case too.

mabrar21:
The code now handles the case where user navigates away without saving changes, reverting to last saved theme. Kudos to @anshumanv  for pointing out the peculiar case.

PR #1139, opened 2018-03-07
Fix issue with code blocks and title on blog page

### LINKED ISSUE #1112: Code blocks on blog page do not fit the page for longer line of codes
**Actual Behaviour**
codeblock

**Expected Behaviour**
It shoud be within the page.

**Steps to reproduce it**
Breaking those lines to fit the size or make those boxes responsive.

Yes.

### LINKED ISSUE #1117: Width of first div of blog page is shorter than width of page
**Actual Behaviour**
div

**Expected Behaviour**
Width should be relative not absolute.

**Screenshots of the issue**
div2

Yes.

PrP-11:
Fixes #1112 
Fixes #1117 
Changes:Made code blocks width responsive 

Demo Link: http://pale-plant.surge.sh/

Screenshots for the change: 
Before:
codeblocks

Now:
codeblocks1

### DIFF
--- a/src/components/Blog/Blog.css
+++ b/src/components/Blog/Blog.css
@@ -154,6 +154,13 @@ pre{
     word-wrap: break-word;
     white-space: pre;
     font-size: 16px;
+	display :block;
+}
+table{
+	display : block;
+	overflow-x: auto;
+	white-space: nowrap;
+	
 }
 @media only screen and (min-width: 1400px){
     .blog_navigation {


PrP-11:
@hkedia321 I didn't got your point, can you explain more what the problem is?

hkedia321:
@PrP-11 when you ask susi queries like `recipe list for chicken` it shows a table. Due to the changes you made, a horizontal scroll is introduced in that table also. See the screenshot above. As this is an unwanted change in this context, please resolve that. You can use css class selectors to select tables only on the blog page.

src/components/Blog/Blog.css
```diff
@@ -154,6 +154,13 @@ pre{
     word-wrap: break-word;
     white-space: pre;
     font-size: 16px;
+	display :block;
+}
+table{
```
rishiraj824: @PrP-11 Please use classes or specific attributes. Make sure you are not changing anything else in the application.

PR #1152, opened 2018-03-12
Improve ChatApp UI

### LINKED ISSUE #1151: Improve the UI of ChatApp
**Actual Behaviour**

I think the current UI of ChatApp doesn't look really good and can be improved.

**Expected Behaviour**

I have some ideas to improve it and make it look better.

Yes

DNS-404:
Fixes #1151

Changes: Improved the UI of ChatApp.

Demo Link: http://faint-month.surge.sh/

Screenshots for the change: 

Before -
screen shot 2018-03-12 at 7 19 35 pm

After -
screen shot 2018-03-12 at 7 19 51 pm

### DIFF
--- a/src/components/ChatApp/ChatApp.css
+++ b/src/components/ChatApp/ChatApp.css
@@ -253,6 +253,7 @@ header{
 }
 .message-container{
   position: relative;
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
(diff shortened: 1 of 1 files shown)


duttaditya18:
Looks good otherwise!

DNS-404:
@duttaditya18 Thanks for the suggestion. I'll look into it. 😃

tabesin:
Guys, the first one is a bubble, the second a rounded rectangle. In principal we could support many different themes, I don't see that as an improvement but worth to develop as an option.

PR #1153, opened 2018-03-12
Default message for empty responses. Fix #1150

### LINKED ISSUE #1150: Susi waits indefinitely to reply to non sensical strings
**Actual Behaviour**
When a nonsensical string e.g random string of words such as "asjasdbv,sadjh" or a legitimate string with a typo such as "comeon" is presented to the chat server, it takes definitely to reply. By examining the reply it can be seen that no answer is returned in the json file.

**Expected Behaviour**
The server should intelligently reply to typos and in the least provide some feedback to the user that it does not understand the question/string.

**Steps to reproduce it**

Send a random string or a typo ( i have tested with the string "comeon") to Susi.

**Screenshots of the issue**

Where-ever possible attach a screenshot of the issue.

Yes

mabrar21:
Added a default message that is displayed when response of
query is empty, i.e answer was not generated (processed).

Fixes #1150 

Changes: IN  src/actions/API.actions.js :  Added default error message to 
server as response when and empty response is received from server.

Demo Link: http://susi-defaultmsg.surge.sh/

Screenshots for the change: 
image

### DIFF
--- a/src/actions/API.actions.js
+++ b/src/actions/API.actions.js
@@ -12,6 +12,7 @@ const cookies = new Cookies();
 let ActionTypes = ChatConstants.ActionTypes;
 let _Location = null;
 let offlineMessage = null;
+let defaultMessage = 'Sorry, I could not understand what you just said.';
 
 // Handle offline, online events
 window.addEventListener('offline', handleOffline.bind(this));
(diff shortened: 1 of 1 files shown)


tabesin:
The technical question is what type of functionality you implement by the client and what by the server. I would suggest as a rule that you only implement "catch response" functionality in the client when it is also available/implemented already on the server side. The problem here is that you have a single message that will always be rendered when you enter garbage.

PR #1195, opened 2018-05-09
Removes the by default keyboard focus on the Reset Button

### LINKED ISSUE #1194: By default, Reset button is in focus
**Actual Behaviour**

The Reset Button is by default in focus when we click on the Themes Menu List Item.

**Expected Behaviour**

The Reset Button should not be in focus by default.

**Steps to reproduce it**

Go to https://chat.susi.ai and log in. Now click on Themes option in the menu. You'll see that in the modal box that pops up, the Reset button is in focus by default.

**Screenshots of the issue**

resetbuttoninfocus

Yes

Akshat-Jain:
Fixes #1194 

**Changes:** Changed the by default keyboard focus state of Save and Reset button so that they are not in focus by default.


mariobehling:
Please provide a deployment test link in future (always).

PR #1319, opened 2018-06-10
Enable ticks while sending message on SUSI 

### LINKED ISSUE #1310: Enable Sent Ticks while sending message
**Actual Behaviour**

Currently there is no provision on the chat app, like in case of messenger or whatsapp for ticks (indicating message has been sent)

**Expected Behaviour**
screenshot 123

**Steps to reproduce it**

Check by sending a message

**LogCat for the issue**
NA

**Screenshots of the issue**
Added above

Yes

prateekiiest:
Fixes #1310 

Changes: [Add here what changes were made in this issue and if possible provide links.]

Enables ticks (as seen messages) when sender sends a message (like WhatsApp , or Messenger)

Demo Link: http://private-wood.surge.sh/

Screenshots for the change: 

screenshot 124

### DIFF
--- a/src/components/ChatApp/MessageListItem/helperFunctions.react.js
+++ b/src/components/ChatApp/MessageListItem/helperFunctions.react.js
@@ -45,6 +45,11 @@ export function renderAnchor(text,link){
 export function renderMessageFooter(message,latestMsgID, isLastAction){
 
   let statusIndicator = null;
+  let ticks = (
+    <span>
+     &#9989;
+   </span>
+ );
 
   let footerStyle = {
     display: 'block',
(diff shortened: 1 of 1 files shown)


prateekiiest:
This is actually inspired from a similar feature request on the susi chromebot

prateekiiest:
@DNS-404 like when we first send the message, there is this Clock Icon.
Do we want me to replace that with a single tick, and then replace that with a double tick along with the twitter share icon?

DNS-404:
There's already a clock. 🤔 Then I don't think this is required. Agreed with @Akshat-Jain

prateekiiest:
Ok, I am closing this

prateekiiest:
I thought about implementing the ticks and replacing the clock by single tick and double ticks when seen.

DNS-404:
Please ask mentors before implementing this. If they approve, then go ahead! 😃

prateekiiest:
ok, sure I will post on the gitter channel. I am keeping this open until then :)

mariobehling:
I don't see that this feature makes sense at the moment. Thanks.

Issue #1361, opened 2018-06-20
Implement new features in the Devices tab in settings


Akshat-Jain:
As discussed in the meeting, the following things need to be taken care of in the Devices tab - 

- [x] Remove the 1st column of the table and make Device name as the 1st column
- [x] Improve alignment of the table in the Devices tab to match the styling of other tabs
- [x] Implement "geolocation" and "room" columns in the table in the Devices tab
- [x] Use data from API for the "geolocation" and "room" columns
- [x] Implement remove option in devices tab.
- [x] Implement edit device option in devices tab.
- [x] Implement a confirmation dialog while removing devices
- [ ] Implement 2 sections below "Devices" - Table view and Map view

**Would you like to work on the issue?**

Yes

anshumanv:
@Akshat-Jain don't write `fixes` unless the PR solves the issue completely.

PR #1383, opened 2018-06-24
Update device info on server when Edit icon is used in the Devices table

Akshat-Jain:
Partially solves #1361
- [x] Implement edit device option in devices tab.

Changes: Now, editing info of the device using the Edit icon would save the updated device info on the server.

Demo Link: https://pr-1383-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 
editdevicesonserver


mariobehling:
Thanks!

PR #1384, opened 2018-06-25
Fix issues with mobile tab

### LINKED ISSUE #1378: Mobile tab in settings on working
**Actual Behaviour**
Adding/changing mobile number is not working

**Expected Behaviour**
We should be able to Add/change mobile number.

**Steps to reproduce it**
Try saving a mobile number in settings

Yes

PrP-11:
Fixes #1378 

Changes: 
- Fix issue with adding mobile number to server. Now we can successfully save our mobile to server.
- Change position of mobile tab just below password tab.
- Change the message to "In future, we will text a verification code to your number"

Demo Link: https://pr-1384-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 
s1

### DIFF
--- a/src/actions/API.actions.js
+++ b/src/actions/API.actions.js
@@ -443,11 +443,11 @@ export function getSettings() {
       settings.prefLanguage = settings.PrefLanguage;
       settings.timeZone = settings.TimeZone;
       settings.customThemeValue = settings.ThemeValues;
-      settings.LocalStorage = true;
-      settings.checked = false;
-      settings.CountryCode = 'US';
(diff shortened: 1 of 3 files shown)


src/actions/API.actions.js
```diff
@@ -447,7 +447,7 @@ export function getSettings() {
       settings.checked = false;
       settings.CountryCode = 'US';
       settings.CountryDialCode = '+1';
```
fragm3: Change above 4 settings, should come from Settings.react.js as they are already set default there in UserPreferences.js

PR #1386, opened 2018-06-25
Fixes #1382: Only logged in users should see thumbs up/down

### LINKED ISSUE #1382: Only logged in users should be able to use thumbs up/down
**Actual Behaviour**
Anyone is able to use thumbs up and down.

**Expected Behaviour**
It should be limited to logged in users.

Yes

anshumanv:
Fixes #1382 

Changes: Thumbs up/down should be only visible to logged in users.

Demo Link: https://pr-1386-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 
Logged in - 
image

Logged out - 
image

### DIFF
--- a/src/components/ChatApp/MessageListItem/Feedback.react.js
+++ b/src/components/ChatApp/MessageListItem/Feedback.react.js
@@ -6,6 +6,9 @@ import ShareIcon from 'material-ui/svg-icons/social/share';
 import UserPreferencesStore from '../../../stores/UserPreferencesStore';
 import * as Actions from '../../../actions/';
 import $ from 'jquery';
+import Cookies from 'universal-cookie';
+
+const cookies = new Cookies();
 
 class Feedback extends React.Component {
   // Parse skill meta data
@@ -128,7 +131,7 @@ class Feedback extends React.Component {
     let feedbackButtons = null;
     let feedbackStyle = {
       top: 5,
-      display: 'block',
+      display: 'flex',
       position: 'relative',
       float: 'right',
     };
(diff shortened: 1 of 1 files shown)


mariobehling:
Actually, I am not sure anymore, if this is really what we want. We save logs according to users. So, even that data might be relevant for us.

@Orbiter What is your opinion?

anshumanv:
@mariobehling you also suggested that we should see a message like `please login to provide feedback` instead of thumbs up/down, but that would only clutter each message box so I didn't add it.

AnupKumarPanwar:
@mariobehling @anshumanv I think we need it but we can have a better implementation. Like we can show the thumbs up and thumbs down to every user, but if he/she is not logged in then a pop up should appear asking the user to log in. And regarding whether we need this or not, as we want that the same user should not be able to change the feedback over same interaction, so we require something to identify the user.

anshumanv:
Yes, one way could be to display the login popup once a logged out user click on thumbs up or down or we can let the user stay anonymous and collect data for anonymous users.

anshumanv:
Yes, agreed.

AnupKumarPanwar:
Okay, then I @mariobehling @anshumanv are we continuing with this PR or closing it.

Orbiter:
@mariobehling I have seen weak participation in rating operations in the YaCy project. Therefore having more opportunities to get more votes are better. We actually need a fraud detection to prevent that a vote can be done only once. An alternative to a vote restriction is, that a re-vote for the same skill must delete the previous vote. But that is unrelated here. For now I would consider to close this issue without merging it to maintain the ability that everyone can vote.

anshumanv:
Yes, legit point @Orbiter, therefore, closing this. :+1:

PR #1430, opened 2018-07-08
Fixes #1326: Set cookies for entire SUSI subdomains

### LINKED ISSUE #1326: User can login from different accounts on SUSI web chat and CMS simultaneously
**Actual Behaviour**

Currently, there are following issues :

The user has to login twice if he chooses to switch to the SUSI webchat from the Skill CMS and vice versa.
As the user has to login twice, s/he can use two different accounts to use the chat and Skill CMS simultaneously.

**Expected Behaviour**

This is probably happening because the web chat and Skills CMS are dealt with differently. We need to have a mechanism wherein the user should be asked to login only once and hence should be allowed to use a single account to use both, the chat and the CMS, simultaneously. Also, the user should not be prompted for login twice. If s/he has logged into the SUSI webchat, then s/he should automatically be logged into the Skills CMS with the same account and vice versa.

No, anybody is free to take this up.

akshatnitd:
Fixes #1326 

Changes: [Add here what changes were made in this issue and if possible provide links.]
* Set cookies for entire SUSI subdomains
* Changed the domain attribute of cookies

Demo link not applicable. Changes made are in accordance to fossasia/susi_skill_cms#902

Screenshots for the change:
NA

### DIFF
--- a/src/components/Auth/ChangePassword/ChangePassword.react.js
+++ b/src/components/Auth/ChangePassword/ChangePassword.react.js
@@ -177,8 +177,8 @@ export default class ChangePassword extends Component {
     }
 
     let email = '';
-    if (cookies.get('email')) {
-      email = cookies.get('email');
+    if (cookies.get('emailId')) {
+      email = cookies.get('emailId');
       console.log(email);
     }
     let changePasswordEndPoint =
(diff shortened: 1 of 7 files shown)


akshatnitd:
Please ignore the Codacy error, as it is due to double quotes, and eslint and codacy are conflicting.

PR #1509, opened 2018-08-08
Fixes #1507: Turn the delete account button into warning color

### LINKED ISSUE #1507: Turn the delete account button into warning color
**Actual Behaviour**
Delete account button seems like a normal button which can be clicked accidentally.

**Expected Behaviour**
It should be colored in warning color. I know we follow a generic theme but warning color seems more suitable to me for this action.

**Screenshots of the issue**
image

Yes

anshumanv:
Fixes #1507
Changes: Quick CSS color update to more appropriate button color.

Demo Link: https://pr-1509-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 
image

### DIFF
--- a/src/components/ChatApp/Settings/Settings.css
+++ b/src/components/ChatApp/Settings/Settings.css
@@ -84,7 +84,7 @@
 }
 
 .Link a {
-  color : #1DA1F5;
+  color : #F44336;
   text-decoration : none;
 }
 


mariobehling:
Twitter also does not use a warning color. We should probably have a section "Danger Zone" like Github, but this is lower priority right now. Closing it therefore. Please focus on finalizing issues we talked about earlier.

Issue #1512, opened 2018-08-09
Pages always call duckduckgo API, which provides http


mariobehling:
If RSS function was used in recent history, the RSS function seems to be called on every single page request (even for settings) resulting in console errors. Is there an alternative to Duckduckgo, e.g. a React JS component that reads RSS instead of calling another service?

`jquery.js:9679 Mixed Content: The page at 'https://chat.susi.ai/' was loaded over HTTPS, but requested an insecure script 'http://api.duckduckgo.com/?format=json&q=search%20for%20atomic%20mass%20of%20radium&callback=jQuery3310025533923962986815_1533785040437&_=1533785040440'. This content should also be served over HTTPS.
send @ jquery.js:9679`

PR #1516, opened 2018-08-09
Used https instead of http for duckduckgo API calls

Akshat-Jain:
Ref: #1512 

Changes: Used `https` instead of `http` for duckduckgo API calls

Demo Link: https://pr-1516-fossasia-susi-web-chat.surge.sh


mariobehling:
This is a first step, but the question in the PR is: How can we get rid of duckduckgo for the RSS functionality? What is an alternative without using external services?

PR #1593, opened 2018-09-27
fixed background change in settings-menu when theme gets changed

### LINKED ISSUE #1587: Background behind settings-menu gets changed even when theme is not selected
**Actual Behaviour**

When opted to change the theme, theme preview appears but when switched to another settings, background behind settings menu gets changed.

**Expected Behaviour**

The background behind the settings-menu should not change when switched to another setting.

**Steps to reproduce it**

Initially, the theme was light

screenshot 54

When I change the theme but I don't save it

screenshot 57

Now, when I switch to some other setting, it still displays the background for dark theme
**LogCat for the issue**

**Screenshots of the issue**

Yes

singhrohan62:
Fixes #1587 

Changes: Background colour behind settings-menu doesn't change now if another theme was selected other than current theme.

Demo Link: https://pr-1593-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 
Initially, the theme was light
screenshot 64

Now I selected dark theme radiobutton but didn't saved changes
screenshot 65

Now when I move to other setting, the bakground remains same of light theme
screenshot 66

Same happens in the case of dark theme
screenshot 67

screenshot 68

screenshot 69

### DIFF
--- a/src/components/ChatApp/Settings/Settings.react.js
+++ b/src/components/ChatApp/Settings/Settings.react.js
@@ -772,10 +772,8 @@ class Settings extends Component {
   // Handle change to theme settings
   handleSelectChange = (event, value) => {
     value === 'light' || value === 'custom'
-      ? (document.getElementById('settings-container').style.background =
-          'rgb(242, 242, 242)')
-      : (document.getElementById('settings-container').style.background =
-          'rgb(0,0,18)');
+      ? $('#settings-container').addClass('settings-container-light')
+      : $('#settings-container').addClass('settings-container-dark');
     this.preview = true;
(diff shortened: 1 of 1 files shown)


singhrohan62:
@fragm3 follow the instructions provided in issue #1587 you will see that the background colour of the settings-container gets changed even though you have just selected radio button of another theme option and not saved it.

singhrohan62:
I think you haven't understood the problem, check this gif to clear your vision
issue1587

YogeshSharma0201:
@singhrohan62 the change in background color when you change theme is just for preview, you have to click on 'Save changes' for the changes to persist.

PR #1607, opened 2018-10-05
fix search functionality when testing skills

### LINKED ISSUE #1602: Search functionality is buggy on testing skills via SUSI Skill CMS
**Actual Behaviour**

Search functionality is not working properly on the test example links which open when we try to test skills via SUSI Skill CMS.

**Expected Behaviour**

Search functionality should work properly.

**Screenshots of the issue**

bugsearchfunctionality

No, anyone can claim this issue.

YogeshSharma0201:
Fixes #1602 

Changes: The problem was that when search changes from false to true, the messageComposer component re-renders which then triggers creation of a new message which happens in its componentDidMount function. When new message is created search is set from true to false, which again renders the messageComposer component and the message is created again. 

The messageComposer component re-renders when search state changes because of this line https://github.com/fossasia/chat.susi.ai/blob/master/src/components/ChatApp/MessageSection/MessageSection.react.js#L687

If we see the ternary statement on this line, the code for when the search is true and when its false is almost same and their is no need to re-render the messageComposer component whenever search state changes. Removing duplicate code solves the problem.

Demo Link: https://pr-1607-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 
ezgif-2-2e3456d08609

### DIFF
--- a/src/components/ChatApp/MessageSection/MessageSection.react.js
+++ b/src/components/ChatApp/MessageSection/MessageSection.react.js
@@ -684,76 +684,79 @@ class MessageSection extends Component {
               searchState={this.state.searchState}
             />
           </header>
-          {!this.state.search ? (
-            <div>
-              <div className="message-pane">
-                <div className="message-section">
-                  <ul
-                    className="message-list"
-                    ref={c => {
-                      this.messageList = c;
+
+          <div>
+            <div className="message-pane">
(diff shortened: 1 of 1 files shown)


Akshat-Jain:
@anshumanv @DNS-404 Please review so that this can be merged if everything's fine. Thanks :-)

PR #1636, opened 2018-10-21
Fixes field text color  in password tab (default custom theme)

### LINKED ISSUE #1635: Field names not visible in password section when selecting custom theme (default settings)
**Actual Behaviour**

Currently when we set custom theme , in password section , text is not visible.(it is set white on white background)

**Expected Behaviour**

Text should be visible in custom theme by default.(ie it should be black probably)

**Steps to reproduce it**

Login -> User Settings -> Set custom theme -> Save -> Click password Tab to view

**Screenshots of the issue**

bug custom-theme

Yes

mkfeuhrer:
** Fixes #1635 **

Changes: Updated color check for custom theme in src/components/Auth/ChangePassword/ChangePassword.react.js

Demo Link: https://pr-1636-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 

updatedcolorbug

### DIFF
--- a/src/components/Auth/ChangePassword/ChangePassword.react.js
+++ b/src/components/Auth/ChangePassword/ChangePassword.react.js
@@ -280,7 +280,11 @@ export default class ChangePassword extends Component {
       minWidth: '150px',
       float: 'left',
       marginTop: '12px',
-      color: UserPreferencesStore.getTheme() === 'light' ? 'black' : 'white',
+      color:
+        UserPreferencesStore.getTheme() === 'light' ||
+        UserPreferencesStore.getTheme() === 'custom'
+          ? 'black'
+          : 'white',
     };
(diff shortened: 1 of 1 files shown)


Akshat-Jain:
@akshatnitd We would have to use a color complimentary to the background color set according to the custom theme? 🤔

mkfeuhrer:
@akshatnitd This is for default custom settings where background is light , if user is setting custom color as dark , then he would accordingly set a contrast color for foreground.

mkfeuhrer:
@akshatnitd From my perspective , PR was to tackle default custom theme case but we can maybe set contrasting colors too. 
Choosing a contrast color depending on background , can you suggest some way on how to tackle this problem ? I can think of in terms of RGB , complimenting each of RGB content.

PR #1653, opened 2018-10-26
Remove legacy String Refs and findDOMNode #1652

### LINKED ISSUE #1652: Remove legacy String Refs and findDOMNode
**Actual Behaviour**

MapContainer contains legacy way of handling with ref by using this.refs. For finding that element, uses findDOMNode

**Expected Behaviour**

This can be avoided using Callback refs.

Yes

fragm3:
Fixes #1652 

Changes: 

- Use callback refs instead of legacy string ref which leads to use of ReactDOM.findDOMNode( )
- Use arrow function in event listener

Demo Link: https://pr-1652-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 
screenshot 2018-10-27 at 2 33 07 am

### DIFF
--- a/src/components/MapContainer/MapContainer.react.js
+++ b/src/components/MapContainer/MapContainer.react.js
@@ -1,5 +1,4 @@
 import React, { Component } from 'react';
-import ReactDOM from 'react-dom';
 // eslint-disable-next-line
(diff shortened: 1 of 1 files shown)


fragm3:
Please ignore codacy error.

PR #1684, opened 2018-11-04
converting all normal functions to arrow functions in utils

### LINKED ISSUE #1683: Refactoring in ./src/utils
**Actual Behaviour**

The current functions in utils use normal JS functions.

**Expected Behaviour**

As we plan to have a more consistent code base, it should be replaced with arrow functions.

Yes

AakashMallik:
Fixes #1683 

Changes: I have converted all the normal functions to arrow functions.

Demo Link: https://pr-1684-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/utils/ChatMessageUtils.js
+++ b/src/utils/ChatMessageUtils.js
@@ -1,13 +1,13 @@
 // Message Utils required for Chat
-export function convertRawMessage(rawMessage, currentThreadID) {
+export const convertRawMessage = (rawMessage, currentThreadID) => {
   return {
     ...rawMessage,
     date: new Date(rawMessage.timestamp),
     isRead: rawMessage.threadID === currentThreadID,
   };
-}
+};
 
-export function getCreatedMessageData(text, currentThreadID, voice) {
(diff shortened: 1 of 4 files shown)


src/utils/index.js
```diff
@@ -1,2 +1,4 @@
-export { default as camelizeKeys } from './camelizeKeys';
-export { default as toFormData } from './toFormData';
+import { camelizeKeys } from './camelizeKeys';
+import { toFormData } from './toFormData';
+
```
akshatnitd: Was change required in this file?

AakashMallik: Sir, I thought it would be more consistent to the rest of the codebase. Should I revert the change?

akshatnitd: Yes, I think you should revert the changes of the file. The current approach of exporting is better,

AakashMallik:
I have made the changes you asked for, sir.

AakashMallik:
@akshatnitd Yes sir, I have changed my commit message to fit the template.

PR #1782, opened 2018-12-02
Initial ActionType and state defined ( messages )

### LINKED ISSUE #1776: Initial redux setup for handling messages ( Reducer, Actions and ActionTypes )
**Actual Behaviour**

Our redux store does not handle messages as of now.

**Expected Behaviour**

The changes I propose are:
- [ ] Define ActionTypes.
- [ ] Add Reducer and initial state.
- [ ] Define Actions.
- [ ] Link actions with API calls.

Yes

AakashMallik:
Fixes #1776 

- [x] Define ActionTypes.
- [x] Add Reducer and initial state.
- [x] Define Actions.
- [x] Define Api calls
- [x] Link actions with API calls.

Demo Link: https://pr-1782-fossasia-susi-web-chat.surge.sh


AakashMallik:
For people reviewing this, here is a brief of things I have done. This is to speed up the review process.
1. Set of actions have been defined
2. Set of reducers have been defined
3. Set of APIs have been defined
4. The message reducer section covers all actions that would be needed to handle messages - User and SUSI.
5. For testing purpose when the **ChatApp** component renders, it calls the **initializeMessageStore()** function which in turn dispatches **MESSAGES_GET_HISTORY_FROM_SERVER**. This will initialize the chat history from the server. When the server responds with history, a **MESSAGES_CREATE_HISTORY_MESSAGES** is dispatched which takes care of filling up the message store accordingly.
6. For testing purpose, when a message is composed in **MessageComposer**, the **createMessage()** is called which in turn dispatches the **MESSAGES_CREATE_USER_MESSAGE**, this adds the message to the store, sends the message to the server and waits for its response. Upon receiving a reply from the server, a **MESSAGES_CREATE_SUSI_MESSAGE** is dispatched, which the reply from SUSI in the store. 
7. For now, feedback is not functional.
8. The messages being properly managed by this new system so far. 

**Note:** Use the redux addon/app to provide me with suggestions and critical advice on how to improve the implementation. The code is still in a very infant stage so any kind of advice would be duly noted.

src/redux/actions/messages.js
```diff
@@ -0,0 +1,215 @@
...
+
+const returnArgumentsFn = function(payload) {
+  return Promise.resolve(payload);
+};
+
+const messageActions = {
+  markAllRead: createAction(actionTypes.MESSAGES_MARK_ALL_READ),
+  createMessage: createAction(
+    actionTypes.MESSAGES_CREATE_USER_MESSAGE,
+    payload => {
```
akshatnitd: Do not do this part here. Use `returnArgumentsFn` here.

src/redux/actions/messages.js
```diff
@@ -0,0 +1,215 @@
...
+      return Promise.resolve(payload);
+    },
+  ),
+  resetVoice: createAction(actionTypes.MESSAGES_RESET_MESSAGE_VOICE),
+  saveFeedback: createAction(
+    actionTypes.MESSAGES_FEEDBACK_RECEIVED,
+    returnArgumentsFn,
+  ),
+  initializeMessageStore: createAction(
+    actionTypes.MESSAGES_GET_HISTORY_FROM_SERVER,
```
akshatnitd: Same for this

src/redux/reducers/messages.js
```diff
@@ -0,0 +1,125 @@
+import { handleActions } from 'redux-actions';
+import actionTypes from '../actionTypes';
+
+const defaultState = {
+  messages: [],
+  messagesByID: {},
+  unreadMessagesByID: {},
```
akshatnitd: `unreadMessageIds: []`

Change the logic accordingly.

src/redux/actions/messages.js
AakashMallik: Sir, I was planning to use ```returnArgumentsFn``` but then I realized that I have to call other actions from within thin action, so I went for this implementation. Where ever I have not used ```returnArgumentsFn``` it is because I had to either call another action or modify the payload before passing it to the reducer. Let me know if there is a more elegant way of doing this that I might be missing out. Thankyou for your time, sir. @akshatnitd

src/redux/reducers/messages.js
AakashMallik: @akshatnitd We won't be able to search for a message by its id in O(1) time if we convert it to an array. Waiting for further clarification from your side.

akshatnitd: eg - 
```
messagesById : {
    1 : {
            "text": "Hi"
    },
    2 : {
            "text": "Hi"
    },
},
unreadMessageIds : [ 2 ],
}
``` 

Therefore info about unreadMessage - 
`messagesById[unreadMessageIds[0]]`

akshatnitd: It can be done in o(1)

src/redux/actions/messages.js
akshatnitd: There is no issue in calling another action from within another in a component.

AakashMallik:
```
createMessage: createAction(
    actionTypes.MESSAGES_CREATE_USER_MESSAGE,
    payload => {
      const timestamp = Date.now();
      const { text, voice } = payload;
      payload.message = {
        id: 'm_' + timestamp,
        authorName: 'You',
        date: new Date(timestamp),
        text: text,
        type: 'message',
        voice: voice,
      };
      apis
        .getSusiReply(payload)
        .then(response => {
          console.log(response);
          store.dispatch(messageActions.createSusiMessage({ response, voice }));
        })
        .catch(error => {
          console.log(error);
        });
      return Promise.resolve(payload);
    },
  ),
```

Sir for example in this code, I am manipulating the payload before passing it on to the api, in this case, if I call the action/ manipulate the payload from the component where this action is called, than the manipulation code and the second action call would need to be repeated every time this action would be called in a component. I hope I haven't got this wrong in my head. Thank you for your detailed review. @akshatnitd

AakashMallik: calling just another action would have been easy to be shifted to the components, but payload manipulation, not so sure.

AakashMallik:
@akshatnitd Also, sir, should be by default put all the incoming message IDs in unreadMessageIDs? and when the user focuses on the chatComponent, we could dispatch the markreadAll action?

src/redux/actions/messages.js
akshatnitd: There is not issue for the manipulation to be done in the component itself. It will give the developer more sense over the actions while being called. For making the component code cleaner, you should make a utils for it ( eg - `formatUserMessage`). And then call another action with the formatted message.

akshatnitd: I see that the formatting for user message and SUSI message is used at multiple places, so would be good to make utils for them.

akshatnitd:
We will be pushing to the unread messages array only if user is out of focus of the window/tab. On focus, we will then mark it read and call appropriate action.

src/redux/actions/messages.js
AakashMallik: @akshatnitd Okay this, I think the new commit has addressed the requested changes. I have got rid of the repetition in the message formatting utility file. I have also removed the payload modification and api calls from actions like you asked.

AakashMallik:
@akshatnitd I am really sorry for taking so much time on this PR. I hope this does not take more than a day or two to finalize. The only this left is the TODO for websearch and rss. If you are okay with the current commit, I'll proceed to finish those TODOs. Thanks a lot for being so patient with my work.

AakashMallik:
@akshatnitd Sir, in that case, we are good to merge. It is just the websearch and rss that will not work. SO empty messages will show up in their place. If you feel like merging this PR, I'll add the rss and websearch in another PR. Your call.

AakashMallik:
@akshatnitd While implementing the web search, I realized that those duckduckgo api calls would end up being promises and I would not be able to call the **createHistoryMessages()** actions without having all the promised in a resolved statel and had to change all the functions to a promise based approach. Now as all of them are promises, so the implementation won't break even if the server responds with a massive latency and we assume these messages to be async. Let me know your thoughts on this, the changes have been made in **formateMessage.js**, **ChatApp.react.js** and **MessageComposer.js** files. 

I was not sure how to trigger the rss action so I am not sure how to test this. Could you provide me with the sample phrase that would trigger the rss feed?

AakashMallik:
Thanks sir.

PR #1797, opened 2018-12-08
Refactor redux bindActionCreator, actionTypes, actions #1796

### LINKED ISSUE #1796: Refactor Redux actions, actionTypes
**Actual Behaviour**

Currently mapDispatchToProps is being used, the code can be simplified and shortened

**Expected Behaviour**

Import only specfic actions, actionType. No need to explicitly dispatching actions and mapping using mapDispatchToProps. 

Yes

fragm3:
Fixes #1796 

Changes: Import only specific actions, actionType. No need to explicitly dispatching actions and mapping using mapDispatchToProps.

Demo Link: https://pr-1796-fossasia-susi-web-chat.surge.sh


src/redux/actionTypes/index.js
```diff
@@ -1,12 +1,8 @@
...
-  'APP_GET_LOGIN',
-  'APP_GET_API_KEYS',
-  'APP_GET_FORGOT_PASSWORD',
-  'SETTINGS_GET_USER_SETTINGS',
-  'SETTINGS_REMOVE_USER_DEVICE',
-  'SETTINGS_ADD_USER_DEVICE',
-  'APP_GET_CHANGE_PASSWORD',
-  'APP_GET_SIGNUP',
-]);
+export const APP_GET_LOGIN = 'APP_GET_LOGIN';
```
AakashMallik: I think the old way of doing it was better. We will have to modify the names twice for every change we make. Chances of introducing an accidental bug would be high in this case. @akshatnitd What do you think?

fragm3:
Will create PR again

PR #1813, opened 2018-12-09
Fixes #1811: Logout action added, force refresh has been removed

### LINKED ISSUE #1811: Create a LOGOUT actionType to clean up the store instead of force reloading.
**Actual Behaviour**

Currently, the app force reloads every time ChatApp is rendered, which makes it impossible for the redux tool to inspect the GET_LOGIN action.

**Expected Behaviour**

Get rid of this force reload, add a LOGOUT actiontype to clean up the message store.

Yes

AakashMallik:
Fixes #1811 

- [x] Added a logout actiontype and respective reducer for it. 
- [x] Call this action whenever the logout component is rendered
- [x] Change the implementation of TopBar to respond dynamically to state change after login/logout with the new redux store.

Demo Link: https://pr-1813-fossasia-susi-web-chat.surge.sh


AakashMallik:
@anshumanv  @akshatnitd Sir, I need some consultation on this PR fossasia/chat.susi.ai#1813 .
So while I was making the TopBar work with the redux store, I realized that the cookies are not getting updated fast enough to be used as a source of truth for the app.
Ex-

cookies.get('emailId')

This will not get updated fast enough to be used in while state transitions from Logged in to Logged out, or vice-a-versa.
The only place we actually need to use cookies is when the app starts for the first time, so I shifted these cookies just to hydrate the initial state of the store. ( I wanted to ask if there was a better way to go about this ).

Then again, we don't update the cookies fast enough, so avatarProps = getAvatarProps(cookies.get('emailId'), accessToken) in topbar.react.js practically throws error as by the time the state changes, the cookie is still undefined. And in our current implementation in Get_LOGIN action, we are not passing the email value to the store, so I have hard coded my email ID for now to test my PR, which seems to be working as I had intended. So now, no force reload will be needed to update the store in any case. Let me know your advice to deal with the problems I have stated above.

Note: People who are well versed with the chat code base can look into this matter too. Head over the my PR and check my current implementation. Thanks

src/components/Auth/Login/Login.react.js
```diff
@@ -106,6 +120,7 @@ class Login extends Component {
           }
         })
         .catch(error => {
+          console.log(error);
```
pythongiant: Is this needed ?

AakashMallik: oh.. nice, you just gave me an idea. I'll make some changes and resubmit the PR.

pythongiant: haha cool

AakashMallik:
@akshatnitd @Akshat-Jain @anshumanv  Let me know if any changes are needed

src/components/Auth/Login/Login.react.js
```diff
@@ -319,7 +319,13 @@ class Login extends Component {
 
 function mapDispatchToProps(dispatch) {
   return {
-    actions: bindActionCreators(actions, dispatch),
+    actions: bindActionCreators(
+      {
+        ...appActions,
+        ...messageActions,
```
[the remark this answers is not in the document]
AakashMallik: Yes sir, this was not needed, I have removed it. My bad.

src/components/ChatApp/ChatApp.react.js
```diff
@@ -11,18 +10,43 @@ import { createMessagePairArray } from '../../utils/formatMessage';
...
+
   componentDidMount() {
+    this.getMessageHistory();
     document.title = 'SUSI.AI Chat - Open Source Artificial Intelligence';
-    // force an update if the URL changes
-    history.listen(() => this.forceUpdate());
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const { uuid, email, userName, accessToken } = nextProps.app;
```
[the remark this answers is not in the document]
AakashMallik: @akshatnitd Sir, I did this to fire this.setState whenever there is a change in app object in the redux store, as the component only re-renders when there is a change in the local state of the component. This would also trigger a rerender for all the children component as you know, and this is what we intend to achieve here as top bar is a child component of the **ChatApp** component. This would also give us a chance to fetch the message store after the login has been completed. I found it to be the most common way of detecting store prop change, so I have used this. Let me know if I missed something.

src/redux/reducers/app.js
```diff
@@ -18,13 +29,23 @@ export default handleActions(
...
+    [actionTypes.APP_LOGOUT](state, { payload }) {
+      return {
+        ...defaultState,
       };
     },
   },
-  defaultState,
+  {
+    ...defaultState,
+    ...cookiesAppValues,
```
fragm3: Can't we use cookiesAppValues directly inside default state?

src/components/ChatApp/ChatApp.react.js
```diff
@@ -11,18 +10,44 @@ import { createMessagePairArray } from '../../utils/formatMessage';
...
+    if (uuid !== this.state.uuid) {
+      this.getMessageHistory();
+      this.setState({
```
fragm3: If this is done through redux, action can be dispatch and the props will change. As the props changes, the view will be re-rendered.

src/components/ChatApp/ChatApp.react.js
```diff
@@ -11,18 +10,44 @@ import { createMessagePairArray } from '../../utils/formatMessage';
 class ChatApp extends Component {
   static propTypes = {
     actions: PropTypes.object,
+    app: PropTypes.object,
   };
 
+  constructor(props) {
+    super(props);
+    const { uuid, email, userName, accessToken } = props.app;
+    this.state = {
```
fragm3: Using props and then assigning it to state, can't we use redux state directly?

src/components/ChatApp/ChatApp.react.js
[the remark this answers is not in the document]
fragm3: @AakashMallik A component can re-render with prop change also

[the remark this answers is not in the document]
AakashMallik: @fragm3 Last time I checked, a component will only re-render if its state changed, or we manually called shouldComponentUbdate().
https://lucybain.com/blog/2017/react-js-when-to-rerender/
If there is some new version update that I am unaware of then do let me know.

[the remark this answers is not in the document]
AakashMallik: @fragm3 You were right, the articles got me confused. Yes indeed props will trigger re-render. The purpose of componetWillRecieve Props is to make sure when to trigger the message fetching. So we only need to fetch the messages when the uuid changes ( Login / Logout ). That is the only purpose of componentWillRecieve props here. Thanks for helping me out here.

src/redux/reducers/app.js
AakashMallik: No, because then the value of default state might look like
{
  uuid : '123433'
}

now when you do something like this,
{
 username: '',
...defaultState,
}

then you would always end up with '123433',
which would be wrong.

I could give you a more elaborate description of this situation.

src/components/ChatApp/ChatApp.react.js
AakashMallik: Yup.. correct, good catch. This will just be needed to make sure that we only fetch new message when there is a change in uuid.

src/redux/reducers/app.js
```diff
@@ -1,15 +1,26 @@
...
+  userName: '',
   email: '',
   accessToken: '',
   uuid: '',
   isAdmin: false,
   apiKeys: {},
 };
 
+const cookiesAppValues = {
+  email: cookies.get('emailId'),
```
akshatnitd: Try using `cookies.getAll` and then assigning individual values. Multiple `get` can be slower.
https://github.com/reactivestack/cookies/tree/master/packages/universal-cookie#getalloptions

src/components/ChatApp/ChatApp.react.js
akshatnitd: `uuid` is not needed here. Incase user logs out, we need to call the action to fetch messages from there itself.

akshatnitd: This may cause unnecessary re-renders.

src/redux/reducers/app.js
AakashMallik: @akshatnitd I have changes the names of cookies to match the state keys, made it easier to destructure from cookies.getAll(). hope it isn't a problem. :sweat_smile:

AakashMallik:
@akshatnitd I have made a storeMessage initialization call in logout and login like you asked. I have removed the previous uuid checking implementation. 
I have renamed a few functions and actionTypes to make it more sensible. 

As ajax.js was using cookies to get access token so I was kind of skeptical about it, so I put a console log to check if it is sending the updated values of access token correctly or not. And yup it is working fine. Let me know if any further changes are needed or if you want me to revert something.

src/components/Auth/Logout.react.js
```diff
@@ -16,24 +19,31 @@ const deleteCookie = function(name, options = {}) {
...
   deleteCookie('emailId', { domain: cookieDomain, path: '/' });
   deleteCookie('username', { domain: cookieDomain, path: '/' });
-  deleteCookie('showAdmin', { domain: cookieDomain, path: '/' });
+  deleteCookie('isAdmin', { domain: cookieDomain, path: '/' });
   deleteCookie('uuid', { domain: cookieDomain, path: '/' });
 
-  // Redirect to landing page
   if (props.history) {
+    props.actions.logout();
     props.history.push('/');
```
AakashMallik: Didn't have to call the **fetchMessage** action because of this statement, as this out unmount the ChatApp and remount it, thus call the **fetchMessage** action from ChatApp's **componentDidMount** lifecycle hook.

AakashMallik:
@akshatnitd Sir I thought @fragm3 was working on the settings reducer so he would be taking care of it. I'll make the changes ASAP.

AakashMallik:
@akshatnitd Made the changes.

AakashMallik:
Btw I have deliberatly removed the logout case from message reducer as everytime the fetch from server is called, it refeshes the who store anyway, so logout would not be necessary there. But in case if you want me to put it back there, i'll add it. Let me know.

AakashMallik:
@akshatnitd Sir by user chat, do you mean we should remove the chat history completely? I could not get you. 
1. Should we wipe out history and not show any chat when a user is logged out? 
2. Or have I made some implementation mistake?

AakashMallik:
Oh... wait.. I get it.. Sir I am working on integrating the messageSection with the redux store. I have an income PR https://github.com/fossasia/chat.susi.ai/pull/1857 sitting. Without that integration the user would be shown the chat for now. I will rush that PR ASAP.

src/redux/reducers/messages.js
```diff
@@ -119,6 +119,11 @@ export default handleActions(
         messagesByID,
       };
     },
+    [actionTypes.APP_LOGOUT](state, { payload }) {
```
AakashMallik: @akshatnitd Not sure if this is what you meant by cleaning app the message store. To be honest this is redundant here but I think for the sake of code consistency, this should work well.

AakashMallik:
@akshatnitd @anshumanv I have probably explained the final changes I have made in a justified way. Let me know if you need further changes. Thank you for your time.

AakashMallik:
@akshatnitd  Yes sir, it would need some fine tuning also some of the reducers I wrote in the messages.js still don't have actions so they probably will end up being redundant.

PR #1824, opened 2018-12-11
Fixes #1823: Snackbar moved to parent component.

### LINKED ISSUE #1823: Move Snackbar to Parent component without breaking functionality
**Actual Behaviour**
The current Snackbar**s** are located in MessageSection.react.js and there are multiple snackbar for each error, which makes them redundant.  

**Expected Behaviour**

A single snackbar must be place at the root, and the function to open it must be passed to the child components.

Yes

AakashMallik:
Fixes #1823 

Changes: The changes I have made in this PR are:
- [x] Moved snackbar to **App.js** which happens to be the parent component.
- [x] Removed redundant instances of Snackbar from MessageSection.react.js
- [x] Passed functions to ChatApp so that its children can open snackbar when needed.
- [x] Place online and offline listeners in App.js and removed them from MessageSection.react.js.

Demo Link: https://pr--fossasia-susi-web-chat.surge.sh

Note: The app did not break after my changes, let me know if anything else is needed.


AakashMallik:
@akshatnitd @anshumanv Can this be reviewed / merged so that I can proceed with this issue https://github.com/fossasia/chat.susi.ai/issues/1853 ?

src/App.js
```diff
@@ -40,9 +41,36 @@ class App extends Component {
...
     };
   }
 
+  componentDidMount = () => {
+    window.addEventListener('offline', this.onUserOffline);
+    window.addEventListener('online', this.onUserOnline);
+  };
+
+  componentWillUnmount = () => {
+    window.addEventListener('offline', this.onUserOffline);
```
akshatnitd: You should remove event listener while unmounting.

src/App.js
```diff
@@ -80,8 +108,39 @@ class App extends Component {
...
+      snackBarDuration,
+      snackBarActionHandler,
+      snackBarAction,
+    });
+  };
+
+  closeSnackBar = () => {
+    this.setState({
+      snackBarOpen: false,
+      message: '',
```
akshatnitd: Also reset the other state variables of snackbar. Otherwise it will cause bugs.

Eg - Suppose you show a snackbar with an action and it auto-closes. Now, even without attaching a snackbar handler to next snackbar messages, the action handler would still be present.

PR #1836, opened 2018-12-12
Add getAdmin in action Creator,reducer,use axios #1808

### LINKED ISSUE #1808: Add getAdmin in action Creator, reducer, use axios for api
**Actual Behaviour**

Currently admin is being set only in local react state using jquery ajax method.

**Expected Behaviour**

Integrate redux for admin state, which would allow users to use admin globally. Also, use axios for api for admin

Yes

fragm3:
Fixes #1808 

Changes:

1. Created getAdmin function to fetch using axios.
2. Added getAdmin to Login.js and removed from StaticAppBar, TopBar
3. Created action creator and reducer for getAdmin
4. Added mockStore for enzyme for StaticAppBar

Demo Link: https://pr-1808-fossasia-susi-web-chat.surge.sh


src/apis/index.js
```diff
@@ -36,6 +38,12 @@ export function getChangePassword(payload) {
   });
 }
 
+export function getAdmin() {
+  const url = `${API_URL}/${AUTH_API_PREFIX}/showAdminService.json}`;
+  // eslint-disable-next-line camelcase
+  return ajax.get(url, { access_token: cookies.get('loggedIn') });
```
AakashMallik: The ajax file has been moded so that it will automatically send this access_token value if available, so you don't need to put the acess_token here.

src/components/Auth/Login/Login.react.js
```diff
@@ -128,6 +128,23 @@ class Login extends Component {
             loading: false,
           });
         });
+      if (cookies.get('loggedIn')) {
```
AakashMallik: Instead of relying on cookie value, you should use the state value. Cookie value are only supposed to be used to hydrate the store. Also, you should shift this code block inside the 
```
if (payload.accepted) {

}
```
block, then you don't even need to check whether he is logged in or not coz if that statement enters the payload. accepted block then technically he is logged in.

src/components/ChatApp/TopBar.react.js
```diff
@@ -90,38 +86,6 @@ class TopBar extends Component {
...
-          console.log(newErrorThrown);
-        },
-      });
-
-      this.setState({
-        showAdmin: cookies.get('showAdmin'),
-      });
-    }
     // Check Logged in
     if (cookies.get('loggedIn')) {
```
AakashMallik: I did not edit this part in my PR because I knew you were working on this block of code. Again, use the state value instead of cookies.

fragm3: @AakashMallik hey, I guess this could be taken look in a new PR. This branch has issue while resolving conflicts. So would be better to take it up in a new PR, anyways this is for getAdmin.

src/components/Auth/Login/Login.react.js
fragm3: Good catch, will make the changes

src/apis/index.js
fragm3: Thanks for telling!

src/components/ChatApp/TopBar.react.js
AakashMallik: @fragm3 I think I can easily make this change in my PR https://github.com/fossasia/chat.susi.ai/pull/1813, as I have taken care of state hydration there and passing the props to TopBar. So if this PR gets merged first, i'll make this change in my PR.

fragm3: Okay great!

src/apis/index.js
```diff
@@ -36,6 +36,11 @@ export function getChangePassword(payload) {
   });
 }
 
+export function getAdmin() {
+  const url = `${API_URL}/${AUTH_API_PREFIX}/showAdminService.json}`;
+  return ajax.get(url);
```
AakashMallik: I think you should put an empty object like the function below this.

src/components/Auth/Login/Login.react.js
```diff
@@ -128,6 +128,23 @@ class Login extends Component {
             loading: false,
           });
         });
+      actions
+        .getAdmin()
+        .then(({ payload }) => {
+          let { isAdmin, accepted } = payload;
+          if (accepted) {
+            cookies.set('showAdmin', true, {
```
AakashMallik: isAdmin

PR #1839, opened 2018-12-13
Fixes #1838: Initial state of settings must be hydrated with cookies [WIP]

### LINKED ISSUE #1838: Hydrate initial state of settings.
**Actual Behaviour**

The initial state of the store is not being hydrated using cookies.

**Expected Behaviour**

The initial state must be hydrated using the save cookies for an anonymous user as it was intended earlier.

Yes

AakashMallik:
Fixes #1838 

* The redux store responds to @INIT action with cookie values passed into default store value.

Demo Link: https://pr-1839-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/redux/reducers/settings.js
+++ b/src/redux/reducers/settings.js
@@ -1,5 +1,8 @@
 import { handleActions } from 'redux-actions';
 import actionTypes from '../actionTypes';
+import Cookies from 'universal-cookie';
+
+const cookies = new Cookies();
 
 const defaultState = {
   theme: 'light',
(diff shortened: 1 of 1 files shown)


AakashMallik:
@akshatnitd Sir, 
```
cookies.set('settings', settings);
```
is being called in Line no. **454** in **Actions.api.js** which we have replaced with the api/index.js file. So cookie section is not visible as of now. But the idea will be to set the cookies everytime the user saves anything in the settings section.

Also, I have fixed the deployment link.  Let me know if you need something else done in this PR.

AakashMallik:
@akshatnitd Sir, judging by the comments the previous developer had left, this cookie was intended to be used for anonymous users and not the ones that are logged in. I guess this is because anonymous users won't have any settings. Also, when the user is logged out, he would technically be an anonymous user, in that case, you won't be able to fetch the settings using API as the app does not have **uuid** for the user.

Let me know if you want to scrap this existing idea and try out something new.

AakashMallik:
This will not be needed anymore. Another PR has dealt with the issue that this PR was addressing.

PR #1845, opened 2018-12-13
Fixes #1827: Blog loading issue fixed

### LINKED ISSUE #1827: Blog TAB is not fetching on a refresh of the page
**Actual Behaviour**

The Blog Tab Fetches the blogs only on switching the tab , but it doesn't work when the page is refreshed or directly visited

**Expected Behaviour**

It should load the blogs even when refreshed or visited directly

**Steps to reproduce it**

- Go to https://chat.susi.ai/blog directly
- or refresh the page after switching to the Blog Tab

**Screenshots of the issue**

I tried in my localhost if the problem was due to any recent pr... but still the same
image

The real hosted site
image

@AakashMallik is working on it

AakashMallik:
Fixes #1827 

* As the API key was outdated, the **count** parameter was not being accepted by the rss2json api, so I removed the count param.

Demo Link: https://pr-1845-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/components/Blog/Blog.react.js
+++ b/src/components/Blog/Blog.react.js
@@ -31,6 +31,7 @@ const loadingStyle = {
   marginTop: '20px',
   position: 'relative',
 };
+
 const allCategories = [
   'FOSSASIA',
   'GSoC',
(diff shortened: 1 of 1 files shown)


AakashMallik:
@isuruAb Thanks for pointing this out, I probably have a better solution to this problem. :+1:

AakashMallik:
@isuruAb I am going to address this problem as part of a broader fix in this issue https://github.com/fossasia/chat.susi.ai/issues/1850. Will drop a PR ASAP.

PR #1806, opened 2018-12-09
Add reducers, actionCreators, actionTypes for Settings#1802

fragm3:
Fixes #1802 
Changes: Added action, reducers for more settings to integrate redux for tabs.
Demo Link: https://pr-1802-fossasia-susi-web-chat.surge.sh

PR #1861, opened 2018-12-18
Add reducers, actionCreators, actionTypes, Set User setting API for Settings#1802

### LINKED ISSUE #1802: Add Action creator, reducers, api for settings components
**Actual Behaviour**

Currently only some api, actionTypes, actionCreator, reducers are present.

**Expected Behaviour**

Add more for redux integration

Yes

fragm3:
Fixes #1802 

Closed: https://github.com/fossasia/chat.susi.ai/pull/1806, as added more changes

Changes: Added action, reducers for more settings to integrate redux for tabs. API integration to set user settings

Demo Link: https://pr-1861-1-fossasia-susi-web-chat.surge.sh


src/redux/actionTypes/index.js
```diff
@@ -17,4 +17,20 @@ export default _.mapKeys([
   'MESSAGES_FEEDBACK_RECEIVED',
   'MESSAGES_INIT_TTS_VOICES',
   'MESSAGES_GET_HISTORY_FROM_SERVER',
+  'SETTINGS_SET_USER_SETTINGS',
```
AakashMallik: I think renaming SETTINGS_SET_USER_SETTINGS to SETTINGS_SAVE_USER_SETTINGS would make more sense?

src/redux/reducers/settings.js
```diff
@@ -16,19 +15,31 @@ const defaultState = {
   userName: '',
   prefLanguage: 'en-US',
   timeZone: 'UTC-02',
-  themeValues: '',
+  customThemeValue: [
```
AakashMallik: Could we use this instead of this array? 
```
{
    header: '#4285f4',
    pane: '#f3f2f4',
    body: '#fff',
    composer: '#f3f2f4',
    textarea: '#fff',
    button: '#4285f4',
    textColor: '#000',
  }
```
As it seems the other components such as MessageSection is using these conventions, plus it will be really tough to decide which color is for what component if you save them like this.

AakashMallik:
@fragm3 Also, I would suggest creating actions **Tabwise**, like "SET_SPEECH_VALUES" and passing an object with to this like
```
{
pitch: '123', rate: '123', etc}
```

src/redux/reducers/settings.js
fragm3: Okay, will make the appropriate changes

src/redux/actionTypes/index.js
fragm3: I think Set(Get is being used) is also fine, isn't it? For 'GET' specific settings, we are using one API for fetching and updating. Creating separate action creators, reducers for it, won't matter much

fragm3:
@AakashMallik It won't matter if we use tabwise or single, we are sending whole data of settings together.

AakashMallik:
> @AakashMallik It won't matter if we use tabwise or single, we are sending whole data of settings together.

Is that so? Say we are in the Accounts tab, you change something there and click save, we would only send the accounts tab data to the redux store so that the store value can change. And as all the action creators are promise based you will initiate a savetoserver action as soon as the redux store gets updated. Not sure why would you send all the settings data to redux store, coz everytime the store updates, the props are updated, so all the components would be re-rendered, which might be unnecessary. 

open to discussion.

src/redux/actionTypes/index.js
AakashMallik: One would be to get the settings from server, which one would be used to send them to server?

fragm3: 'SETTINGS_SET_USER_SETTINGS'

fragm3:
@AakashMallik Splitting reducers and combining them at the end, as the store is common. This would result in the same thing with separate reducers itself. The view will be re-rendered. Reducers are split in the same store when, `take pieces of code that are very long or do many different things, and break them into smaller pieces that are easier to understand`. Link: https://redux.js.org/recipes/structuring-reducers/splitting-reducer-logic.  In our case, the logic is pretty straight forward, and as of now we don't need to split them tab wise.

AakashMallik:
@fragm3 Actually your implementation would work nicely if we happen to link that save button with the action that does the api call, and for each change the user makes, the actions to set the store are called. Okay, this seems fine. :+1:

src/redux/actionTypes/index.js
AakashMallik: I might sound stupid right now, SETTINGS_SET_USER_SETTINGS is to send the settings to the server, what would you do to initialize the store. Coz I see not api that uses this url ```/aaa/listUserSettings.json?``` which is used to fetch user settings from server.

AakashMallik: API.actions.js Line number 424

fragm3: `chat.susi.ai/src/apis/index.js ` line `39` function does that. getUserSettings is used to fetch the user settings.

AakashMallik: Wait, I can't see it in your PR.. oh.. it was probably implemented by Akshat sir. So do we have an action to call that api or do you plan to call it from the component directly? @fragm3

fragm3:
@akshatnitd @anshumanv please review

src/redux/actionTypes/index.js
fragm3: Action

fragm3:
@akshatnitd I will using action creator and reducer like: SETTINGS_SETTINGS_SET_ACCOUNT_SETTINGS, GET specific can be common for all tabs. Is that fine?

src/redux/reducers/settings.js
```diff
@@ -43,6 +55,62 @@ export default handleActions(
...
+        ...state,
+        phoneNo,
+        countryCode,
+        countryDialCode,
+      };
+    },
+    [actionTypes.SETTINGS_SET_CHAT_PREFERENCES_SETTINGS](state, { payload }) {
+      return {
+        ...state,
+        enterAsSend: !state.enterAsSend,
```
akshatnitd: Do not do this. Send true/false from component

fragm3: Okay

src/redux/reducers/settings.js
```diff
@@ -43,6 +55,62 @@ export default handleActions(
...
+      return {
+        ...state,
+        theme,
+        customThemeValue,
+      };
+    },
+    [actionTypes.SETTINGS_SET_MICROPHONE_SETTINGS](state, { payload }) {
+      return {
+        ...state,
+        micInput: !state.micInput,
```
akshatnitd: Same here

fragm3:
@akshatnitd made the changes, please review

PR #1842, opened 2018-12-13
Add getAdmin in action Creator,reducer,use axios #1808

fragm3:
Fixes #1808 

Had to close https://github.com/fossasia/chat.susi.ai/pull/1836, it was not able to detect new changes

Changes:

1. Created getAdmin function to fetch using axios.
2. Added getAdmin to Login.js and removed from StaticAppBar, TopBar
3. Created action creator and reducer for getAdmin
4. Added mockStore for enzyme for StaticAppBar

Demo Link:  https://pr-1842-1-fossasia-susi-web-chat.surge.sh

PR #1877, opened 2018-12-21
Add getAdmin in action Creator,reducer,use axios #1808

### LINKED ISSUE #1808: Add getAdmin in action Creator, reducer, use axios for api

fragm3:
Fixes #1808 

Had to close https://github.com/fossasia/chat.susi.ai/pull/1842, as branch was old, login button was not working

Changes:

1. Created getAdmin function to fetch using axios.
2. Added getAdmin to Login.js and removed from StaticAppBar, TopBar
3. Created action creator and reducer for getAdmin
4. Added mockStore for enzyme for StaticAppBar

Demo Link:  https://pr-1877-1-fossasia-susi-web-chat.surge.sh


akshatnitd:
I debugged it locally. The showAdminService API call is being cancelled. The reason seems to be the unmounting of the component, before the API call could be made. So, make a small change and call `onRequestClose` only after the Admin API has been completed.

fragm3:
@akshatnitd I have added handleDialogClose as a callback to ensure that admin API is completed before the tab closes. About hydrating the store, should I simply use this in default store(defaultState): `isAdmin: cookies.get('isAdmin') || false`

src/components/Auth/Login/Login.react.js
```diff
@@ -148,6 +148,27 @@ class Login extends Component {
...
+          });
+        }
+        this.setState({
+          isAdmin: showAdmin,
+        });
+      })
+      .catch(error => {
+        console.log('Failed to get admin');
+      });
+    this.handleDialogClose();
```
akshatnitd: Put this inside the then of the getAdmin action block.

src/components/Auth/Login/Login.react.js
```diff
@@ -148,6 +148,27 @@ class Login extends Component {
     }
   };
 
+  checkAdmin = () => {
+    actions
```
akshatnitd: This should be `this.props.actions`. Please check the code before submitting.

src/redux/reducers/app.js
```diff
@@ -27,6 +27,13 @@ export default handleActions(
         accessToken,
       };
     },
+    [actionTypes.APP_GET_ADMIN](state, { payload }) {
+      const { isAdmin } = payload;
```
akshatnitd: The API returns `showAdmin` instead of `isAdmin`.

akshatnitd: You can write - `const { showAdmin: isAdmin } = payload` here.

src/components/Auth/Login/Login.react.js
fragm3: Actually actions is already defined, no need to use this.props.actions in this file

fragm3:
The request as you said @akshatnitd of showAdmin is getting cancelled. It is not due to closing of dialog, I have tested that.

fragm3:
@AakashMallik still showing cancelled

AakashMallik:
Well, I'll fork your branch and look into it. I am let you know by 3am.

PR #1957, opened 2019-01-15
Fixes #1947: Workaround for captcha key loading failure

### LINKED ISSUE #1947: Propose a workaround for the case when the captcha fails to load
**Actual Behaviour**

As of now, if the server fails to respond with captcha key, then the captcha doesn't load. But that will also mean that the user won't be able to signup.

**Expected Behaviour**

Propose a workaround for this.

Yes

AakashMallik:
Fixes #1947 

Changes: The changes I have made are as follows:
- [x] The user can signup if the captcha key fails to load.
- [x] User won't be able to signup if the captcha keys loads and he fails to clear the robot test

Demo Link: https://pr-1957-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/components/Auth/SignUp/SignUp.react.js
+++ b/src/components/Auth/SignUp/SignUp.react.js
@@ -283,7 +283,7 @@ class SignUp extends Component {
       !passwordErrorMessage &&
       confirmPassword &&
       !passwordConfirmErrorMessage &&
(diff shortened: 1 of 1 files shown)


AakashMallik:
@anshumanv Lemme know if this is fine by you.

Issue #1953, opened 2019-01-14
Add UI reducer, action, action type for Dialog, SnackBar,UserOnline/Offline etc


fragm3:
**Actual Behaviour**

Currently, in index.js we are passing state and methods to Login, SignUp, ForgotPassword, and etc, down the tree.

**Expected Behaviour**

To handle Dialog, User Status(Online, Offline), Snackbar (#1950 ) which is used globally should be handled using a reducer.

Add appropriate action, action type, reducers

**Would you like to work on the issue?**

Yes

Issue #1950, opened 2019-01-13
Create SnackBar Reducer and Use It Across The App


fragm3:
**Actual Behaviour**

Currently, Snackbar is either being used in Parent component or individually inside every component

**Expected Behaviour**

Create global Snackbar reducer, action, and integrate it across the app

**Would you like to work on the issue?**

Yes

PR #1959, opened 2019-01-16
Add snackbar#1950, Dialog UI#1953 reducer, action, action creator

fragm3:
Fixes: #1953, #1950

Changes:

Add reducers, action, action creator for redux integration for the following:

1. Snackbar
2. Dialog Actions(OpenLogin, OpenSignup, OpenForgotPassword

Demo Link: https://pr-1959-1-fossasia-susi-web-chat.surge.sh


src/redux/actions/snackbar.js
```diff
@@ -0,0 +1,25 @@
+import { handleActions } from 'redux-actions';
+import actionTypes from '../actionTypes';
```
akshatnitd: Do need a separate file for this. It can be combined with the dialog box reducers.

src/redux/actions/dialog.js
```diff
@@ -0,0 +1,37 @@
+import { handleActions } from 'redux-actions';
+import actionTypes from '../actionTypes';
+
+const defaultState = {
+  openLogin: false,
```
akshatnitd: Keep the state as - 
```
modalProps: {
isModalOpen : false,
modalType: ''
},
snackBarProps : {
isSnackBarOpen: false,
snackBarMessage: ''
}
```

src/redux/actions/dialog.js
```diff
@@ -0,0 +1,37 @@
...
+  openSignUp: false,
+  openForgotPassword: false,
+};
+
+export default handleActions(
+  {
+    [actionTypes.UI_OPEN_LOGIN_DIALOG](state) {
```
akshatnitd: Make changes in the actions and reducers accordingly.

src/redux/actions/snackbar.js
fragm3: @akshatnitd The reason for keeping a separate file is, snackbar is used in more components, where dialog box is not used. Just to keep the imports less, so I split the reducers in to 2. If you want, I can combine them, or we can use 2 separate reducer

fragm3: I'll go with your approach, and make a single file

fragm3:
@akshatnitd made the changes 👍 , please review

PR #1974, opened 2019-01-21
Fixes #1973: Words suggestion for input field

### LINKED ISSUE #1973: Words suggestion for input field
**Actual Behaviour**

Words suggestion is not implemented for the input in the text box. 

**Expected Behaviour**

While typing in the text box, words should be suggested as in google keyboard.

**Screenshots of the issue**

issue

Yes

AvinashAgarwal14:
Fixes #1973 

Changes: Implemented words suggestion for the input field.

Demo Link: https://pr-1974-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 

sugg


fragm3:
@AvinashAgarwal14 the idea is good, but it's supposed to suggest relevant words(little relevant also works). It would be great, can you try to implement that?

AvinashAgarwal14:
Can you give me few examples where you see that it is not suggesting relevant words ?
It might happen for some cases because of lack of data. Currently it has only 1000 words to compare the input string. But it does show correct suggestion in most cases.

fragm3:
@AvinashAgarwal14 I can see only random words popping. Wait for other reviewers comment

AvinashAgarwal14:
@fragm3 Okay. Let me explain how this is working.
For example: 
In case of incomplete search words: If you type 'wh' then 'while', 'white', 'who', 'whole', 'why' comes as suggestions (according to alphabetical order). Only most commonly used words are suggested.

In case of a complete search word: Suppose you type 'name' since 'name' is a word in itself and there are very few commonly used words that go with 'name', the first suggestion is 'name' itself, while the other four suggestion depends upon previous matches - 'nam' or 'na'. The other four less accurate suggestions are not removed inorder to make the UI consistent.

src/components/ChatApp/MessageListItem/generateMessageBubbleUtil.js
```diff
@@ -364,52 +365,52 @@ export const generateMessageBubble = (
...
-                mapAnchor,
-                mymap,
-                message,
-                latestUserMsgID,
-                showFeedback,
-              ),
-            );
-            break;
-          }
+          // case 'map': {
```
AakashMallik: May I know the reason for commenting this code out?

src/components/ChatApp/MessageListItem/generateMessageBubbleUtil.js
```diff
@@ -106,29 +106,29 @@ const generateAnchorBubble = (
...
-        <div className="message-text">{replacedText}</div>
-        <div>{mapAnchor}</div>
-        <br />
-        <div>{mymap}</div>
-        {renderMessageFooter(message, latestUserMsgID, showFeedback)}
-      </section>
-    </div>
-  );
-};
+// const generateMapBubble = (
```
AakashMallik: Justify

src/components/ChatApp/Settings/PreviewThemeChat.js
```diff
@@ -3,43 +3,80 @@ import PropTypes from 'prop-types';
 import { getSusiPreviewReply } from '../../../apis/index';
 import susiWhite from '../../../images/susi-logo-white.png';
 import MoreVertIcon from 'material-ui/svg-icons/navigation/more-vert';
+import loadingGIF from '../../../images/loading.gif';
```
AakashMallik: Is this even related to your issue?

src/components/ChatApp/Settings/Settings.react.js
```diff
@@ -213,7 +213,7 @@ class Settings extends Component {
         console.log(errorThrown);
       },
       complete: function(jqXHR, textStatus) {
-        location.reload();
+        window.location.reload();
```
AakashMallik: location prop is passed by the router, not sure why you have replaced them with window.location

src/components/ChatApp/words.json
```diff
@@ -0,0 +1,985 @@
+{
```
AakashMallik: To be honest, this don't really give a personlaized recomendation based on user typing patterns, thus making the whole feature reduntant. This feature will need more discussion to be even considered "required".

src/components/StaticAppBar/StaticAppBar.react.js
```diff
@@ -275,7 +273,7 @@ class StaticAppBar extends Component {
     const TopRightMenu = props => {
       let avatarProps = null;
       if (accessToken) {
-        avatarProps = getAvatarProps(cookies.get('emailId'));
+        avatarProps = getAvatarProps(email);
```
AakashMallik: Why are these changes done? They are not related to your issue.

src/utils/invertColor.js
```diff
@@ -1,4 +1,4 @@
-export const invertColorTextArea = textarea => {
+export const invertColor = textarea => {
```
AakashMallik: Revert this change. This function is being used in other PRs and this will cause unnecessary merge conflicts. If need be, the change would be made by the person dealing with that PR.

src/components/ChatApp/words.json
AvinashAgarwal14: Yes, you are right. It does not give results based on user typing pattern. It is a very basic feature in which the last word of the input stream is compared with an array of 1000 commonly used words so that users does not need to type entire words.

AvinashAgarwal14:
All the above snippets of code were commented because I was not able to build the application otherwise.
The image below shows the errors that were thrown.

error

I don't know why do they occur, maybe because of the difference of node/npm or other packages version.

fragm3:
Closing, out of scope.

PR #2005, opened 2019-01-27
Fixes #1945: User now notified if emailID is already taken

### LINKED ISSUE #1945: Handle errors for existing emailID for Signup api calls.
**Actual Behaviour**

The error messages are not descriptive after the refactoring process.

**Expected Behaviour**

Make the error cases more descriptive to help the user better understand the problem with the failed api call.

**Steps to reproduce it**
- EmailID taken and network error throw similar error messages.

yes

AakashMallik:
Fixes #1945 

- [x] If email ID is already taken then the user will be given a different error message. Making the error more expressive will improve UX.

Demo Link: https://pr-2005-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/components/Auth/SignUp/SignUp.react.js
+++ b/src/components/Auth/SignUp/SignUp.react.js
@@ -244,14 +244,20 @@ class SignUp extends Component {
           }
         })
         .catch(error => {
-          console.log(error);
           this.setState({
             success: false,
             password: '',
             loading: false,
           });
+          let snackBarMessage;
+          if (error.statusCode === 422) {
+            snackBarMessage =
+              'Already registered. Please signup with a different email account';
+          } else {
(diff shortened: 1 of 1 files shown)


src/components/Auth/SignUp/SignUp.react.js
```diff
@@ -244,14 +244,19 @@ class SignUp extends Component {
...
         .catch(error => {
-          console.log(error);
           this.setState({
             success: false,
             password: '',
             loading: false,
           });
+          let snackBarMessage;
+          if (error.statusCode === 422) {
+            snackBarMessage = 'Email ID not available';
```
akshatnitd: Please use a different message here - `Already registered. Please signup with a different email account.`

AakashMallik:
@akshatnitd Made the change you asked for.

PR #2012, opened 2019-01-29
Added Signup option in navigation bar.

plxity:
Fixes #2010 

Changes: Added Sign Up option in Top Bar

Demo Link: https://pr-2012-fossasia-susi-web-chat.surge.sh

Screenshots for the change: 
capture

### DIFF
--- a/src/components/ChatApp/TopBar.react.js
+++ b/src/components/ChatApp/TopBar.react.js
@@ -10,6 +10,7 @@ import Popover from 'material-ui/Popover';
 import Settings from 'material-ui/svg-icons/action/settings';
 import Exit from 'material-ui/svg-icons/action/exit-to-app';
 import SignUp from 'material-ui/svg-icons/action/account-circle';
+import PersonAdd from 'material-ui/svg-icons/social/person-add';
 import Info from 'material-ui/svg-icons/action/info';
 import Dashboard from 'material-ui/svg-icons/action/dashboard';
 import List from 'material-ui/svg-icons/action/list';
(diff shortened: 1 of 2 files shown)


plxity:
@anshumanv @akshatnitd Please review this. :)

src/components/ChatApp/TopBar.react.js
```diff
@@ -256,6 +259,16 @@ class TopBar extends Component {
               onTouchTap={toggleShareClose}
               rightIcon={<Share />}
             />
+            {accessToken ? (
```
akshatnitd: Use `&&` inplace of ternary here

anshumanv:
Not sure if we need this as this will break the consistency across all apps, so we gotta do it everywhere in case we decide to do it here.

plxity:
@anshumanv I feel adding this option will provide smooth functioning for new users to signup. and in https://accounts.susi.ai/ I think we don't need this as homepage already has an option for signup. We need to add this only in https://skills.susi.ai/. If you approve this, I will create a similar issue in other repo.

src/components/ChatApp/TopBar.react.js
plxity: Okay, sir.

plxity:
@anshumanv What should I do?

anshumanv:
@akshatnitd what do you think about the change? 🤔

plxity:
@fragm3 but why?

plxity:
@akshatnitd What do you suggest or think about this feature? :)

plxity:
@akshatmitd Okay, but if someone is new to this application and wants to sign-up, only "Login" option will appear in the navigation bar. It will be confusing. But if Sign-up option is present it will provide a direct link to Sign-up.
That's all from my side :)

plxity:
@mariobehling Sir, What do you think about this feature? :)

plxity:
@mariobehling Sir, What are your views on this?

fragm3:
Not required, closing.

PR #2031, opened 2019-02-07
Fixes #2006: User now notified in real time if an email ID is already taken

### LINKED ISSUE #2006: Add a feature to check if an email ID is available in real time while user enters the emailID
**Actual Behaviour**

As of now, the user has to enter the email id and all the other details and press the submit button to see if the emailID is actually available. 

**Expected Behaviour**

A feature should be introduced so that this process can be done in real time and thus the user would know if the email ID is available as soon as he fills the email ID field. 

Yes

AakashMallik:
Fixes #2006

- [x] User now notified in real time if email ID is already taken
- [x] I made use of a debounced function to minimize api call clutter.

Demo Link: https://pr-2031-fossasia-susi-web-chat.surge.sh

Screenshot:
screenshot-localhost-3000-2019 02 07-14-45-20

### DIFF
--- a/src/apis/index.js
+++ b/src/apis/index.js
@@ -142,3 +142,11 @@ export function getAdmin(payload) {
   const url = `${API_URL}/${AUTH_API_PREFIX}/showAdminService.json`;
   return ajax.get(url, payload);
 }
+
+export function getEmailExists(payload) {
+  const { email } = payload;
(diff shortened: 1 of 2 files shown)


AakashMallik:
Thought of doing that but Akshat  sir made me change the error message in a previous PR so I thought of using the same... I also prefer shorter messages but... :shrug:

plxity:
oh okay, no issues then 👍

PR #2053, opened 2019-02-17
Tightly couple verify password and new password

### LINKED ISSUE #2049: tightly couple verify password and new password
**Actual Behaviour**

When something is entered in the fields and the removed, the validation text remains.

**Expected Behaviour**

The validation text should not be displayed when the fields are empty.

**Steps to reproduce it**

1. Settings -> Password
2. Type something in any of the 3 inputs requested.
3. Clear the input.

**Screenshots of the issue**

issue7

Yes

ArexJambusarwala:
Fixes #2049 

Changes: Changes in file ChangePassword.react.js are made which when input is cleared no error message is shown.

Demo Link : https://pr-2053-fossasia-susi-web-chat.surge.sh/

Screenshots of the change: 

corrected6

### DIFF
--- a/src/components/Auth/ChangePassword/ChangePassword.react.js
+++ b/src/components/Auth/ChangePassword/ChangePassword.react.js
@@ -167,8 +167,8 @@ class ChangePassword extends Component {
 
         const newPasswordConfirmError =
           ((confirmNewPassword || newPasswordConfirmErrorMessage) &&
-            !(confirmNewPassword === newPassword)) ||
-          (newPassword === '' && newPasswordConfirmErrorMessage);
+          !  (confirmNewPassword === newPassword)) ||
+         (newPassword === '' && newPasswordConfirmErrorMessage);
 
         this.setState({
           newPassword,
--- a/src/components/Auth/ChangePassword/ChangePassword.react.js
+++ b/src/components/Auth/ChangePassword/ChangePassword.react.js
@@ -167,7 +167,8 @@ class ChangePassword extends Component {
 
         const newPasswordConfirmError =
           ((confirmNewPassword || newPasswordConfirmErrorMessage) &&
-          !(confirmNewPassword === newPassword)) || (newPassword === '' && newPasswordConfirmErrorMessage);
+          !  (confirmNewPassword === newPassword)) ||
+         (newPassword === '' && newPasswordConfirmErrorMessage);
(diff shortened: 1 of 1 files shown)


ArexJambusarwala:
If the error message is supposed to be shown, that it has some inconsistencies. For example, if we type something in verify password and remove it, the error message is displayed. Now if we type something in "new password" field and remove it,  then error message of "verify password" goes away.(sice both fields now have same value)

AakashMallik:
> If the error message is supposed to be shown, that it has some inconsistencies. For example, if we type something in verify password and remove it, the error message is displayed. Now if we type something in "new password" field and remove it, then error message of "verify password" goes away.(since both fields now have same value). I will fix the unsuccessful check if this change is indeed required.

The verify password issue is a legit case. I had raised an issue regarding this but that was done for the signup component. You can follow along with my old PR and make similar changes here. https://github.com/fossasia/chat.susi.ai/pull/1767
I coupled the verify password and password field together to address this problem, but it a **similar** problem but from a different **perspective**. Let me know if you need further clarification.

ArexJambusarwala:
I hope this is what is required:
http://pr-2053-fossasia-susi-web-chat.surge.sh/

Verify password error message remains unlike before.

AakashMallik:
CI failing.. please check plus files changed turn up to be zero...

ArexJambusarwala:
The Cl test fails stating:
   Aborted - you do not have permission to publish to pr-2053-fossasia-susi-web-chat.surge.sh
How to go through with this?

src/components/Auth/ChangePassword/ChangePassword.react.js
```diff
@@ -166,8 +166,9 @@ class ChangePassword extends Component {
           : '';
 
         const newPasswordConfirmError =
-          (confirmNewPassword || newPasswordConfirmErrorMessage) &&
-          !(confirmNewPassword === newPassword);
+          ((confirmNewPassword || newPasswordConfirmErrorMessage) &&
+            !(confirmNewPassword === newPassword)) ||
+          (newPassword === '' && newPasswordConfirmErrorMessage);
```
AakashMallik: replace ```newPassword === ''``` with ```!newPassword```

fragm3:
Closing, PR not needed

PR #2062, opened 2019-03-05
Codes Optimized & README Updated

Shehanka:
Changes: [Readme file updated, I've added License to the file.]

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,4 +260,4 @@ chat.susi.ai/
 
 ## License
 
-This repository is under a GNU LESSER GENERAL PUBLIC LICENSE 2.1.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -260,4 +260,4 @@ chat.susi.ai/
 
 ## License
 
-This repository is under a GNU LESSER GENERAL PUBLIC LICENSE 2.1.
+This repository is under a GNU LESSER GENERAL PUBLIC LICENSE 2.1, see the [LICENSE.md](https://github.com/fossasia/chat.susi.ai/blob/master/LICENSE) file for details
```
AdityaSrivast: @Shehanka it would be better to use only (LICENSE) instead of the longer link

Shehanka: That link is won't show up in view mode it's only visible LICENSE.md part only check out markdown cheatsheet [Link](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links)
like above link what I mentioned

PR #2072, opened 2019-04-13
Remove Save option where Toggle is given in Settings

### LINKED ISSUE #2071: Remove Save option where Toggle is given in Settings
**Actual Behaviour**
Currently the user has to toggle the option and then click on save.

**Expected Behaviour**
On toggle itself the state change should be saved.

**Screenshots of the issue**
Screenshot 2019-04-14 at 1 26 33 AM

I am working on it.

harsh-jindal:
Fixes #2071 

Changes: [Add here what changes were made in this pull request and if possible provide links showcasing the changes.]
Remove Save option where Toggle is given in Settings
Demo Link: https://pr-2072-fossasia-susi-web-chat.surge.sh

Screenshot 2019-04-14 at 2 10 50 AM

### DIFF
--- a/src/components/ChatApp/Settings/Settings.react.js
+++ b/src/components/ChatApp/Settings/Settings.react.js
@@ -726,6 +726,9 @@ class Settings extends Component {
     this.setState({
       enterAsSend: isInputChecked,
     });
+    setTimeout(() => {
+      this.handleSubmit();
+    }, 500);
   };
 
   // Handle change to mic input settings
@@ -733,6 +736,9 @@ class Settings extends Component {
     this.setState({
       micInput: isInputChecked,
     });
+    setTimeout(() => {
+      this.handleSubmit();
+    }, 500);
   };
 
   // Handle change to speech output on speech input settings
@@ -985,6 +991,12 @@ class Settings extends Component {
       return false;
     }
     if (selectedSetting === 'Share on Social media') {
+      return false;
+    }
+    if (selectedSetting === 'ChatApp') {
+      return false;
+    }
+    if (selectedSetting === 'Microphone') {
       return false;
     }
     return true; // display the button otherwise


rossoskull:
I think the users should be given one more chance to finalize their changes, so we should keep the 'Save changes' button. @akshatnitd @Akshat-Jain What are your views?

src/components/ChatApp/Settings/Settings.react.js
```diff
@@ -726,13 +726,19 @@ class Settings extends Component {
     this.setState({
       enterAsSend: isInputChecked,
     });
+    setTimeout(() => {
```
fragm3: Why setTimeout? You can simply call it in callback function. Always avoid using setTimeout

src/components/ChatApp/Settings/Settings.react.js
```diff
@@ -726,13 +726,19 @@ class Settings extends Component {
...
+      this.handleSubmit();
```
fragm3: Why setTimeout? You can simply call it in callback function. Always avoid using setTimeout

src/components/ChatApp/Settings/Settings.react.js
```diff
@@ -726,13 +726,19 @@ class Settings extends Component {
...
+    }, 500);
   };
 
   // Handle change to mic input settings
   handleMicInput = (event, isInputChecked) => {
     this.setState({
       micInput: isInputChecked,
     });
+    setTimeout(() => {
```
fragm3: Same here.

src/components/ChatApp/Settings/Settings.react.js
```diff
@@ -987,6 +993,12 @@ class Settings extends Component {
     if (selectedSetting === 'Share on Social media') {
       return false;
     }
+    if (selectedSetting === 'ChatApp') {
```
fragm3: You can reduce the number of statements by using `||` operator

harsh-jindal:
@akshatnitd @rossoskull   If the user toggles the mic input and does not save there. Instead goes to speech and makes some changes there and click on save, then also the mic option gets saved as we have a common save function. So either we have to implement save function specific to that tab or save on toggle to avoid this.

harsh-jindal:
@akshatnitd My idea was to save the user choice on toggle itself. So the save button is not required on tabs where we just have a toggle button.

akshatnitd:
But, it should be uniform across all the tabs. So, IMO this PR is not required at the first place.

harsh-jindal:
@akshatnitd Okay. But shouldn't we implement separate save functions for each tab. Like for example the user goes and toggles the mic input but does not save there. And then he goes to make some change in speech and clicks on save there. Now his change in mic input also gets saved. So shall I open a separate issue for that? I'll close this PR.

PR #2144, opened 2019-05-31
[FIX] : Fix getDefaultMapData API function

### LINKED ISSUE #2126: [API] : API not defined for getDefaultMapData
**Actual Behaviour**

Should call API as defined

**Steps to reproduce**
Type in chat : 'where am I' and see the resultant screen as code breaks

**Screenshots of the issue**

Screenshot from 2019-05-24 11-46-24

Yes, but provide necessary business logic behind it

meetzaveri:
Fixes #2126 

Changes: Change in action for getDefaultMapData for user for query `where am I`

Demo Link: https://pr-2144-fossasia-susi-web-chat.surge.sh

Screenshots of the change: 
Screenshot from 2019-05-31 12-24-56


meetzaveri:
@felixonmars Actually when you said in gitter conversation, that we have now redux integrated into project as your were migrating actions from that, so I thought you were indirectly conveying that I should use a redux action for this.

Problem is that: While using simple API, that's an asynchronous task and I tried it before moving to redux action. I was not getting expected result through it as there was no way coming back because due to no change in state or props in parent component, the component wasn't re-render.

I achieved this via implementing redux action as it was only way to update store hence re-rendering the component.

fragm3:
@meetzaveri for asynchronous we use promise, you can wait for an async task to finish and put the logic in then block.

meetzaveri:
I exactly tried that way only. But as you know it's a for loop. For loop is synchronous in nature

fragm3:
Okay, let's keep it through redux

src/redux/reducers/messages.js
```diff
@@ -21,6 +22,19 @@ export default handleActions(
...
       };
     },
+    [actionTypes.MESSAGES_GET_USER_GEO_DATA](state, { payload }) {
+      return {
+        ...state,
+        userGeoData: {
+          /* Manual schema stiching to provide necessary fields that are required */
+          ...payload,
+          text: 'Your location',
+          link: `https://www.openstreetmap.org/#map=13/${payload.lat}/${
```
fragm3: We only require `lon, lat`, remove everything else from`userGeoData`

src/components/ChatApp/MessageListItem/generateMessageBubbleUtil.js
```diff
@@ -504,6 +500,7 @@ export const generateMessageBubble = (
         locale = ttsLanguage;
       }
 
+      console.log('Before Return JSX');
```
fragm3: Remove

src/components/ChatApp/MessageListItem/generateMessageBubbleUtil.js
```diff
@@ -4,7 +4,7 @@ import Emojify from 'react-emojione';
 import TextHighlight from 'react-text-highlight';
 import VoicePlayer from './VoicePlayer';
 import YouTube from 'react-youtube';
-import { getDefaultMapData } from '../../../apis';
+// import { getDefaultMapData } from '../../../apis';
```
fragm3: Remove

src/redux/reducers/messages.js
meetzaveri: But we need text and link to display above the card content. For eg. as `here is a map` should point to map link. That's why I did schema stitching there.

src/components/ChatApp/MessageListItem/generateMessageBubbleUtil.js
meetzaveri: Done.

src/components/ChatApp/MessageListItem/generateMessageBubbleUtil.js
meetzaveri: Done

src/redux/reducers/messages.js
fragm3: Create the link and text in the place you are using it. Not in the reducer.

meetzaveri: Okay.

meetzaveri: Have pushed final commit @fragm3

akshatnitd:
Please squash so that we can go ahead and merge it.

meetzaveri:
I squashed them into one commit already. Still there's something remaining?

fragm3:
@meetzaveri make sure you do `git push -f`

meetzaveri:
@akshatnitd @fragm3 Squashed my commits into one and also took pull from upstream master. Please review

meetzaveri:
My latest commit of `[CONFLICT]` is not shown when I do `git rebase -i HEAD~3`. So in order to squash into one, what should I do now ?

PR #2151, opened 2019-06-01
Fix Category wise display of cards in mobile view

### LINKED ISSUE #2148: Fix Category wise display of cards in mobile view
**Actual Behaviour**
Currently the mobile view is broken with a lot of space between the heading and the cards.

**Expected Behaviour**
Fix the mobile view.

**Steps to reproduce it**
Click on any category from sidebar.
Open console. Click on any mobile layout. See error.

**Screenshots of the issue**
Screenshot 2019-05-29 at 8 59 48 AM

Yes

harsh-jindal:
Fixes #2148 
Changes: Fix Category wise display of cards in mobile view

Demo Link: https://pr-2151-fossasia-susi-web-chat.surge.sh

Screenshot 2019-06-01 at 10 35 28 PM

### DIFF
--- a/src/components/cms/BrowseSkill/BrowseSkill.js
+++ b/src/components/cms/BrowseSkill/BrowseSkill.js
@@ -6,7 +6,7 @@ import _ from 'lodash';
 import { Link as _Link } from 'react-router-dom';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 import skillActions from '../../../redux/actions/skills';
 import FormControl from '@material-ui/core/FormControl';
 import InputLabel from '@material-ui/core/InputLabel';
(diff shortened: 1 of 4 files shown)


src/components/cms/BrowseSkill/BrowseSkill.js
```diff
@@ -809,41 +816,76 @@ class BrowseSkill extends React.Component {
...
+                        item
+                        sm={6}
+                        alignContent="flex-end"
+                        style={{ alignItems: isMobile ? 'center' : 'left' }}
+                      >
+                        {skills.length > 10 && (
+                          <div>
+                            <FormControl
+                              style={{
+                                width: isMobileView(350) ? '120px' : '145px',
```
akshatnitd: Why does isMobileView take in a param. It should do so.

fragm3:
@akshatnitd We can do one thing, for all CSS specific changes we can integrate them together with `styled components` and `rem` unit together

src/components/cms/BrowseSkill/BrowseSkill.js
harsh-jindal: @akshatnitd This was done because iphone 5 has 320px width while it works perfectly fine for all phones above 350px width.

harsh-jindal:
@fragm3 Yup I think this can be done. But I guess the only problem would be if we have many components like many same divs with different styles we need to discuss on the nomenclature.

akshatnitd:
@fragm3 Yes, ideally we should do that. We should use CSS to achieve responsiveness. Javascript APIs of innerWidth should be used only if we need some change in JS logic. Like maybe, replacing Checkbox with Radio button based on device.

fragm3:
@akshatnitd @harsh-jindal yes exactly, that's where styled-components will help. We should go with styled components for mobile responsive(talking about its implementation in `FormControl`. For replacing prop related logic or any js related logic we can create a styled components for the same instead of innerWidth.

mariobehling:
Looks good to me, but yes would be great to have the approach outlined above. @fragm3 Could you open an issue, please? 

@akshatnitd I think this can be merged as a step into the right direction, right?

fragm3:
@akshatnitd what can be done for this PR, I will be working on fixing the BrowseSkill, skill by language, etc. https://github.com/fossasia/susi.ai/issues/2219

harsh-jindal:
@akshatnitd Shall I make a seperate variable and call isMobileView(350) so i'll be calling it once ?

harsh-jindal:
@akshatnitd Please review. I have made changes and now used css instead of isMobileView()

PR #2159, opened 2019-06-02
Create AppShell layout, Reuse Footer, StaticAppBar #2158

### LINKED ISSUE #2158: Create AppShell layout
**Actual Behaviour**

Not present

**Expected Behaviour**

Implement AppShell layout consisting of Header and Footer, as said by @akshatnitd 

Yes

fragm3:
Fixes #2158

Changes: Reuse Footer, StaticAppBar

Demo Link: https://pr-2159-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -35,6 +35,8 @@ import SkillHistory from './components/cms/SkillHistory/SkillHistory';
 import SkillRollBack from './components/cms/SkillRollBack/SkillRollBack';
 import SkillCreator from './components/cms/SkillCreator/SkillCreator';
 import BotBuilderWrap from './components/cms/BotBuilder/BotBuilderWrap';
+import StaticAppBar from './components/StaticAppBar/StaticAppBar.react';
+import Footer from './components/Footer/Footer.react';
 import Admin from './components/Admin/Admin';
 
 class App extends Component {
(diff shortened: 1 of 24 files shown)


src/App.js
```diff
@@ -93,6 +99,7 @@ class App extends Component {
             message={snackBarMessage}
             onClose={actions.closeSnackBar}
           />
+          {pathname !== '/' && <StaticAppBar />}
```
akshatnitd: Keep this logic inside StaticAppBar, would be more cleaner.

src/App.js
```diff
@@ -157,6 +164,7 @@ class App extends Component {
             <ProtectedRoute exact path="/settings" component={Settings} />
             <Route exact path="/*:path(error-404|)" component={NotFound} />
           </Switch>
+          {pathname !== '/' || (skillListRegex.test(pathname) && <Footer />)}
```
akshatnitd: Same here.

src/App.js
fragm3: @akshatnitd Yes, but instead of rendering the component and returning it null, this might be a better way.

akshatnitd: I agree. But, the difference is negligible. So, let's go for cleaner code 💯

akshatnitd: Now, it looks fine. But, once there would be multiple logic. We will need to anyways do it.

fragm3: @akshatnitd I tried what you are trying to say, after implmenting it, it is problematic keeping the current code in mind. If I put the logic in StaticAppBar, I am getting this error:

`Uncaught TypeError: Cannot read property 'offsetHeight' of undefined
    at StaticAppBar.componentDidMount (StaticAppBar.react.js:217)`

Screenshot 2019-06-03 at 7 44 31 PM

Also, if we don't return anything, we will have to nest it into React.Fragment which is not desirable either.

fragm3:
@akshatnitd I have used another approach to render the elements, please review.

Issue #2259, opened 2019-06-11
Remove IconButton on User not logged in App Bars


fragm3:
**Actual Behaviour**

A small button is present in StaticAppBar, TopAppBar when user not logged in.

**Expected Behaviour**

Fix it

**Screenshots of the issue**

<img width="95" alt="Screenshot 2019-06-12 at 3 19 07 AM" src="https://user-images.githubusercontent.com/18073126/59309355-f85d8a00-8cc0-11e9-88b5-31d0531042d0.png">

**Would you like to work on the issue?**

Yes

PR #2260, opened 2019-06-11
Fix Cookies policy,  Remove IconButton on User not logged in App Bar

### LINKED ISSUE #2257: Cookies policy is appearing everytime.
**Actual Behaviour**

Cookies policy is appearing at the bottom, every time the webpage loads.

**Expected Behaviour**

It should be shown to the user until the user accepts the policy.

**Steps to reproduce it**

1. Visit https://susi.ai/.
2. Accept the policy.
3. Reload the page.

**Screenshots of the issue**

image

No, anyone can take this.

fragm3:
Fixes #2257, #2259

Changes: 
- [x] Fix small button present in StaticAppBar, TopAppBar when user not logged in.
- [x] Fix Cookies policy appearing at the bottom on page refresh, every time the webpage loads.

Demo Link: https://pr-2260-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -49,6 +49,7 @@ class App extends Component {
     snackBarProps: PropTypes.object,
     showCookiePolicy: PropTypes.bool,
     modalProps: PropTypes.object,
+    visited: PropTypes.bool,
   };
 
   componentDidMount = () => {
@@ -92,6 +93,7 @@ class App extends Component {
       modalProps: { isModalOpen },
       location: { pathname },
       showCookiePolicy,
+      visited,
     } = this.props;
     const skillListRegex = new RegExp('^/skills');
     const pathLength = pathname.split('/').length;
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -110,8 +112,9 @@ class App extends Component {
       renderFooterPagesList.includes(pathname) ? (
         <Footer />
       ) : null;
-    const renderCookiePolicy = showCookiePolicy ? <CookiePolicy /> : null;
-    const renderDialog = isModalOpen ? <DialogSection /> : null;
+    const renderCookiePolicy =
+      showCookiePolicy === true ? <CookiePolicy /> : null;
```
akshatnitd: Why is `=== true` required, if showCookiePolicy is already boolean?

src/App.js
```diff
@@ -110,8 +112,9 @@ class App extends Component {
...
+    const renderDialog = isModalOpen || !visited ? <DialogSection /> : null;
```
akshatnitd: Why have you used `|| !visited`? Won't it open DialogSection if `visited` is false? If you intend to open the Dialog, then where is the type of Dialog defined?

fragm3: The video tour dialog will not open if we don't use this condition, it opens when visited is set to false in `DialogSection`.

src/App.js
fragm3: It is not working with
`const renderCookiePolicy = showCookiePolicy ? <CookiePolicy /> : null`

PR #2263, opened 2019-06-12
[FIX] : Fix customThemeValue property undefined error with setting de…

### LINKED ISSUE #2240: Settings reducer code breaks due to unavaibility of theme property
**Actual Behaviour**
When user is logged in and on screen load, there is issue settings reducer as code breaks. The `customThemeValue` from settings reducer is always undefined as it is not shown in `settings` payload object

**Expected Behaviour**

Check if `customThemeValue` property is available from `settings` payload object, then use it or else use default `customThemeValue` property from `defaultState` object. That way code wouldn't break.

**Steps to reproduce it**

- Direct login with user
- Upon landing on initial overview page, just wait for 10 secs.
- Check network tab in chrome devtools

**LogCat for the issue**

**Screenshots of the issue**

_Issue screenshot says it's `theme` property due to which code breaks, but actually it's `customThemeValue` property._

Screenshot from 2019-06-10 10-14-49
Screenshot from 2019-06-10 10-14-44
Screenshot from 2019-06-10 10-39-55

Yes

meetzaveri:
…fault state

Fixes #2240 

Changes: Fixes the customThemeValue property error of undefined by setting default state for it if it is undefined otherwise takes payload's data

Demo Link: https://pr-2263-fossasia-susi-web-chat.surge.sh


meetzaveri:
@fragm3  Okay I agree, so then what if other keys than `theme` are missing from payload ? Should I make them redundant also ?

meetzaveri:
> Retrigger the Travis build, rest LGTM

How to?

src/redux/reducers/settings.js
```diff
@@ -65,13 +65,16 @@ export default handleActions(
         avatarType,
       } = settings;
       let { customThemeValue } = settings;
-      const themeArray = customThemeValue.split(',').map(value => `#${value}`);
+      let settingsTheme = theme ? theme : defaultState.theme;
```
akshatnitd: Directly write `theme = defaultState.theme` while destructuring settings.

meetzaveri: Okay, on it

meetzaveri:
> There are some extra changes being reflected on your PR. Please remove it

@akshatnitd  Do you mean extra commits from other authors squashed into one ? Can you give example/reference ?

akshatnitd:
Yes. You should squash only your commits.

Eg - `import Settings.css`

meetzaveri:
@akshatnitd Done!

PR #2387, opened 2019-06-27
Add Loader in missing places #2299

### LINKED ISSUE #2299: Add Loader in missing places
**Actual Behaviour**

 Loader missing in 

- Settings component
- Admin Component

**Expected Behaviour**

Add it in missing places

Yes

fragm3:
Fixes #2299

Changes: 

- Add loader System Settings Config Table
- Fix Skill Tab Skills loading used to show `no records to display`
- Removed redundant handlers in skills tab

Demo Link: https://pr-2387-fossasia-susi-web-chat.surge.sh

Screenshots of the change:

Screenshot 2019-06-30 at 12 54 55 AM


fragm3:
@akshatnitd @harsh-jindal 

I have the following questions, please address them:

- Use material table for config keys? Gives a nice loading effect and would be consistent with other tables. 
- Add confirmation of Delete Key, Confirm update, addition?
- Add loading in button in skills tab in admin

harsh-jindal:
@fragm3 We can use material table for config keys if in the future we have many keys and we want to have search options.
Confirmation Delete key dialog is already present for the config key. 
Also confirmation create i think is not required as admin only has the write to create keys.

fragm3:
@harsh-jindal confirmation for delete key is not present

fragm3:
@akshatnitd I can either hide the table to show the Loader on center or we can use this alignment, can't be done inside Table

harsh-jindal:
@fragm3 While deleting key we can have the admin enter key name before deleting as a confirmation. I don't think we need to have two dialog boxes for deletion of config key.

fragm3:
@harsh-jindal I didn't add 2 Dialog

harsh-jindal:
@fragm3 I didn't mean you have added two dialogs. I meant we can have the admin enter the key name before deletion as a confirmation. Can you please tell your idea? Maybe I'm missing your meaning of `confirmation of Delete Key`

fragm3:
@harsh-jindal I thought, as we have less number of keys, we don't need confirmations. Should I add confirmation @akshatnitd like harsh is saying? or let it be?

fragm3:
Updated:
Screenshot 2019-06-30 at 12 37 42 PM

akshatnitd:
@harsh-jindal @fragm3 Confirmation is good idea. Can be handled in a separate PR. 
We should ideally make the confirmation dialog a shared component. It is being used with same design at multiple places.

Issue #2359, opened 2019-06-23
Integrate Styled components in Chat app


samagragupta:
**Actual Behaviour**

Not present.

**Expected Behaviour**

Create integrated styled components.

**Would you like to work on the issue?**

yes

fragm3:
@samagragupta try picking some easy task at first. Thank you!

PR #2397, opened 2019-06-28
Fix color of skills page, cleanup

### LINKED ISSUE #2396: Fix color of skills page
**Actual Behaviour**

The background color is messed up if we go from `/chat` to `/skills`

**Expected Behaviour**

Fix the color

**Steps to reproduce it**

Go to https://susi.ai/chat, then go to https://susi.ai/skills

**Screenshots of the issue**

Screenshot 2019-06-28 at 10 06 21 PM

Yes

fragm3:
Fixes #2396, #2359 

Changes: 
- Fix color of the skills page
Note: I tried using other ways to achieve body color, didn't find any other cleaner way of doing it.

- Add styled components, remove redundant styles
Note: I will be combining TopAppBar with StaticAppBar next PR, so left some CSS styles.

Demo Link: https://pr-2397-fossasia-susi-web-chat.surge.sh

Screenshot:

Screenshot 2019-06-29 at 3 41 24 AM

### DIFF
--- a/src/components/ChatApp/MessageSection/MessageSection.react.js
+++ b/src/components/ChatApp/MessageSection/MessageSection.react.js
@@ -474,6 +474,10 @@ class MessageSection extends Component {
     return LoadingComponent;
   };
 
+  componentWillUnmount() {
+    document.body.style.setProperty('background-color', 'white');
+  }
+
   render() {
     const {
       dream,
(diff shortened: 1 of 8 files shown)


src/components/ChatApp/MessageSection/MessageSection.react.js
```diff
@@ -474,6 +474,10 @@ class MessageSection extends Component {
     return LoadingComponent;
   };
 
+  componentWillUnmount() {
+    document.body.style.setProperty('background-color', 'white');
```
akshatnitd: I didn't understand the approach. How will it work?

fragm3: @akshatnitd The problem is of higher specificity, I tried &&& to make it more specific and use styled-component, didn't work. The ChatApp styles on the body get removed if we unset background-color which solves the problem.

fragm3:
@akshatnitd resolved.

PR #2633, opened 2019-07-18
Option to stop a skill playing audio #2614

### LINKED ISSUE #2614: Options to stop a skill, reduce the volume of playing music
When the web client uses the sound output, there is no way to stop the output. How could we achieve this and stop the sound output in regards to:
* stop music from playing
* stop reading out long articles

How to use the stop or pause command?

fragm3:
Fixes #2614

Changes: Stop currently playing audio from skill using 'stop', 'pause'  or 'abort' voice command

Demo Link: https://pr-2633-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/components/ChatApp/MessageListItem/MessageListItem.react.js
+++ b/src/components/ChatApp/MessageListItem/MessageListItem.react.js
@@ -18,6 +18,7 @@ class MessageListItem extends React.Component {
     actions: PropTypes.object,
     userGeoData: PropTypes.object,
     showChatBubble: PropTypes.bool,
+    pauseAllVideos: PropTypes.func,
   };
 
   constructor(props) {
@@ -96,6 +97,7 @@ class MessageListItem extends React.Component {
       this.onTextToSpeechEnd,
       this.onYouTubePlayerReady,
       this.getUserGeoData,
+      this.props.pauseAllVideos,
     );
   }
 }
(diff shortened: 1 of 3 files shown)


mariobehling:
@fragm3 I think we should follow the approach to use a hotword. What would be needed to make this possible on the webclient?

fragm3:
@mariobehling definitely hotword approach could be used across the web client for more use cases like `okay susi` and it would open up a chat on the right side. But this approach will definitely take a longer time, I'll look into this in more details, if we find the relevant packages, this might be possible and would be a super amazing feature

fragm3:
@akshatnitd But the problem here would be, to stop playing the audio, we would be first sending the response to the backend (how? using speech detection or typing), wait for the response, and then stop. This would result in a delay and be an additional overhead.

akshatnitd:
Yes, I am aware of that. Check gitter channel for discussion. But it will help us implement cross platform solution

fragm3:
@akshatnitd I am finished with skill chat audio stop, removed the global state to see if audio is playing as Orbiter said chat window should always be listening. Updated changes are in 2nd commit. If you give a heads up, I'll rebase

fragm3:
@akshatnitd Hotword detection part needs more discussion. I need more details and what all we require, We need to follow a more structured approach of how to tackle big issues.

For this PR:
1. First commit only listens if susi speech is audio playing the message. 
2. Second commit always listens to user input as said by @Orbiter 

You can test it by typing `how to make biryani` and say stop.

fragm3:
@akshatnitd done, 3rd commits does precisely 
Screenshot 2019-07-19 at 4 16 47 PM

fragm3:
@akshatnitd fixed

akshatnitd:
@fragm3 How do we pause the SoundCloud instances as well?
It should be done in this PR itself.

fragm3:
@akshatnitd soundcloud already pausing, test it in deployed link

akshatnitd:
@fragm3 Mention about this feature on the gitter channel. Good work.

PR #2637, opened 2019-07-20
All the styled components in a single file of NavigationBar

### LINKED ISSUE #2431: All the styled components in a single file
**Actual Behaviour**

In Navigation Bar component styles.js file is created but all the style are not present in that.

**Expected Behaviour**

Shift all the styled components in styles.js

Yes

samagragupta:
Fixes #2431

Changes: Shift all the styled-components in styles.js

Demo Link: https://pr-2637-fossasia-susi-web-chat.surge.sh


samagragupta:
@fragm3 It would be more structured. If we have already created style.js, then why not shift all the styles in that component.

Issue #2640, opened 2019-07-20
Fix email autofill in searchbar


fragm3:
**Actual Behaviour**

If we go to the password tab in settings, the email address gets auto-filled in the search bar

**Expected Behaviour**

Fix it

**Steps to reproduce it**

Go to settings --> Password tab 

**Screenshots of the issue**

<img width="980" alt="Screenshot 2019-07-20 at 6 04 51 PM" src="https://user-images.githubusercontent.com/18073126/61578754-ee794300-ab18-11e9-99bd-790634b4ea37.png">

**Would you like to work on the issue?**

Yes

Issue #2638, opened 2019-07-20
Fix avatar upload user interaction


fragm3:
**Actual Behaviour**

Currently, when editing an avatar, can be done only by edit button. Clicking on avatar image multiple times, lead to error, of image undefined

**Expected Behaviour**

Edit image should open on clicking the uploaded image

**Would you like to work on the issue?**

Yes

Issue #2555, opened 2019-07-15
Only allow usernames with up to 51 characters


mariobehling:
Someone put in a username with hundreds of characters. This could potentially be used by hackers. Please allow only usernames up to 51 characters (also open a related issue on the server).

This policy is similar to Facebook.

https://stackoverflow.com/questions/8078939/what-is-the-maximum-length-of-a-facebook-name

fragm3:
@mariobehling this will be handled in https://github.com/fossasia/susi_server/issues/1306. Do tell more points that come up regarding this

PR #2661, opened 2019-07-23
Fix validation #2546,#2555, searchbar bug #2640, avatar UI #2638

### LINKED ISSUE #2546: Login and Password reset: Improve user security
The system is not hardened against brute-force attacks. There a few small things we can do. Please open relevant issues in the server as well. 

* Require user passwords to have at least 8 characters and at least one special character, one number and text characters
* Add Recaptcha for second log-in attempts
* Add Recaptcha for password resets (forgot password)

Question: How can this be made working with the server? If someone could just call the API from another domain we don't gain anything.

Also, this needs to work with the Android app. Please collaborate with the Android team.

fragm3:
Fixes #2546, #2638, #2555, #2640

Changes: 
- [x] Avatar on click interaction
- [x]  Require user passwords to have at least 8 characters and at least one special character, one number and text characters
 - [x] Add Recaptcha for second log-in attempts
 - [x] Add Recaptcha for password resets (forgot password)
- [x] Username less than or equal to 51 characters.
- [x] Fix email autofill in the search bar

Demo Link: https://pr-2639-fossasia-susi-web-chat.surge.sh

Screenshots of the change:

Screenshot 2019-07-20 at 6 20 01 PM

### DIFF
--- a/src/components/Auth/ChangePassword/ChangePassword.react.js
+++ b/src/components/Auth/ChangePassword/ChangePassword.react.js
@@ -14,13 +14,15 @@ import Translate from '../../Translate/Translate.react';
 import appActions from '../../../redux/actions/app';
 import uiActions from '../../../redux/actions/ui';
 import PasswordStrengthBar from '../../shared/PasswordStrengthBar';
+import isPassword from '../../../utils/isPassword';
+import Recaptcha from '../../shared/Recaptcha';
 
 const PasswordField = styled(_PasswordField)`
   height: 35px;
   border-radius: 4;
   border: 1px solid #ced4da;
   padding: 0px 12px;
-  width: 17rem;
+  width: 22.5rem;
   @media (max-width: 520px) {
     width: 100%;
     max-width: 17rem;
@@ -37,7 +39,7 @@ const ForgotPasswordLink = styled.div`
   margin: 1rem 0;
   color: #1da1f5;
   cursor: pointer;
-
+  width: fit-content;
   a {
     text-decoration: none;
   }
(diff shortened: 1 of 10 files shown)


fragm3:
Note: I have intentionally left the comment in password change as for now users are following old regex and if we change it, they won't be able to change password.

src/components/Auth/ChangePassword/ChangePassword.react.js
```diff
@@ -104,18 +124,15 @@ class ChangePassword extends Component {
     switch (event.target.name) {
       case 'password': {
         const password = event.target.value.trim();
-        const passwordError = !(
-          password &&
-          password.length >= 6 &&
-          password.length <= 64
-        );
+        // const passwordError = !passwordRegEx.test(password)
```
akshatnitd: Do we really require regex in current password? I think we can remove it.

src/components/Settings/AccountTab.react.js
```diff
@@ -22,6 +22,7 @@ import styled from 'styled-components';
 import { setUserSettings, uploadAvatar } from '../../apis';
 import defaultAvatar from '../../images/defaultAvatar.png';
 import CircularProgress from '@material-ui/core/CircularProgress';
+import { userNameRegEx } from '../../utils/regex';
```
akshatnitd: Can you use this constant wherever applicable to remove redundacy? Like for eg - Signup modal.
Same for passwordRegex.

fragm3:
@akshatnitd made the changes

PR #2687, opened 2019-07-27
Scroll bar does not disappear

### LINKED ISSUE #2530: Clicking on "Create" on the left side it results in disappearing the scroll bar of the site
When the user clicks on the blue create button, the scroll page on the right side of the screen disappears.

Screenshot from 2019-07-12 23-20-14

samagragupta:
Fixes #2530 

Changes: When clicked on create (or any other drop-down menu) the scroll bar disappears. 

Demo Link: https://pr-2687-fossasia-susi-web-chat.surge.sh

Or we can use this: https://material-ui.com/components/selects/#native-select


samagragupta:
Or we can use this: https://material-ui.com/components/selects/#native-select

harsh-jindal:
Your current implementation doesn't look like a good solution to the problem. Because if you click on create and then scroll the menu list also scrolls down.
Screenshot 2019-07-28 at 6 49 55 AM

tanujvyas10:
@samagragupta Sir,the language selection icon is also distorted when i open your surge link.
langDis

samagragupta:
So could we can use this: https://material-ui.com/components/selects/#native-select

harsh-jindal:
@samagragupta How will that solve the problem. Even there you cannot scroll the window with select menu open.

samagragupta:
@harsh-jindal Why there is a need to scroll the window when the select menu is open?

harsh-jindal:
@samagragupta The screen shall not scroll. That's my point, then this PR should not be there in the first place. I dont understand what are trying to solve then ?

samagragupta:
@harsh-jindal So according to the issue I have to show the scrollbar.

harsh-jindal:
@samagragupta Okay. Make sure the current bug in you PR link i.e. menu list scrolling down is fixed.

samagragupta:
I will think of any other idea and then will create a new PR.

PR #2689, opened 2019-07-27
Remove BrowseSkillByCategory, BrowseSkillByLanguage

### LINKED ISSUE #2688: Remove BrowseSkillByCategory, BrowseSkillByLanguage
**Actual Behaviour**

Not required

**Expected Behaviour**

Use Route parameters to know about the route type

Yes

fragm3:
Fixes #2688

Changes: Removed BrowseSkillByCategory, BrowseSkillByLanguage

Demo Link: https://pr-2689-fossasia-susi-web-chat.surge.sh


fragm3:
@akshatnitd oh missed that, we are exposing the language routes in skill details, will handle those cases as well.

fragm3:
@akshatnitd If we require the Language pages, then maybe current code is better. Should I undo changes related to route params and use separate pages, let it stay like the current version or use a single BrowseSkill component instead of 3?

PR #2720, opened 2019-07-29
Fix Email Admin Settings#2710

### LINKED ISSUE #2710: Email Config not working, No test emails, Form Issues
* Clicking on dropdown form elements sidebar disappears
Screenshot from 2019-07-29 09-51-47
* SMTP Info should be below SMTP radio button
Screenshot from 2019-07-29 09-51-21
* There is no password box for SMTP mails. How could emails be sent without password?
* Test Email
   * Test Email does not work. Email is not received.
   * There is no success or unsuccessful message for "Test Mail" after clicking button. Has an email been sent or not?
* Mail Type does not make sense. Mail type is set to "Disable". Does this mean no emails are send out? Probably it should be "Mail Gateway" here. Which gateway should the system use? STMP, Sendgrid, or "Disable" for deactivating emails.
Screenshot from 2019-07-29 09-55-05
* The space of the white content area to the left-hand side and to the menu itself is not consistent across "settings subtabs" (please click on the subtabs to compare)
Screenshot from 2019-07-29 10-05-42

fragm3:
Fixes #2710, #2721 

Changes: 
- Fix select popover across the app
- Fix Alignment of white spaces
- Rename mail type to mail gateway
- Fixed test email functionality

Demo Link: https://pr-2720-fossasia-susi-web-chat.surge.sh

Screenshots of the change:

Screenshot 2019-08-03 at 3 20 29 AM

Screenshot 2019-08-03 at 3 24 10 AM

Screenshot 2019-08-03 at 3 24 47 AM


fragm3:
@mariobehling I need the email config's you were using from the actual server instance to make it work.

mariobehling:
> @mariobehling I need the email config's you were using from the actual server instance to make it work.

Please test this with a test account e.g. a Gmail account.

Otherwise you will need to check if test emails arrived, if the password is correct, if the port is wrong etc. Just go ahead with a test account and we will add a standard config afterwards.

fragm3:
@mariobehling I have added test settings and created a test Gmail account, still can't send a test email. The changes on the backend are not changed, I am using the same logic to send a test email, as done in case of forgot password.

akshatnitd:
@fragm3 Where have you added the settings?

fragm3:
@akshatnitd added the settings, now test mail working. Please review.

mariobehling:
I don't see the changes made, that were requested. Were is the password? You did not move the smtp area to the right location below the STMP button etc.

Screenshot from 2019-08-05 11-49-03

Reopening the issue therefore.

PR #2804, opened 2019-08-08
use correct host address for SUSI Sever API URL[WIP]

### LINKED ISSUE #2769: Local-pages build: Server not reachable outside the host machine
https://github.com/fossasia/susi.ai/blob/4d3b8985a4ac9f76c38f52beb9d17246b4f33977/deploy.sh#L91
When opening the WebClient on any machine on the same network it is unable to connect to the SUSI-Server as it tries to connect to a server running on that machine instead of the host.

alok760:
Fixes #2769

Changes: Add a function which checks if `LOCAL_ENV` variable is `true` and use the correct API URL accordingly

Demo Link: https://pr-2804-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/utils/urls.js
+++ b/src/utils/urls.js
@@ -1,5 +1,12 @@
+function check() {
+  if (process && process.env.REACT_APP_LOCAL_ENV === 'true') {
+    return 'http://' + window.location.hostname + ':4000';
+  }
+  return 'https://api.susi.ai';
+}
+
 const urls = {
-  API_URL: process.env.REACT_APP_API_URL || 'https://api.susi.ai',
+  API_URL: check(),
   SOUND_SERVER_API_URL: 'http://0.0.0.0:7070',
   CHAT_URL: 'https://susi.ai/chat',
   SKILL_URL: 'https://susi.ai/skills',


src/utils/urls.js
```diff
@@ -1,5 +1,12 @@
+function check() {
+  if (process.env.REACT_APP_LOCAL_ENV === 'true') {
+    return 'http://' + window.location.hostname + ':4000';
```
akshatnitd: Wouldn't writing `0.0.0.0:4000` be the same?

src/utils/urls.js
```diff
@@ -1,5 +1,12 @@
+function check() {
+  if (process.env.REACT_APP_LOCAL_ENV === 'true') {
```
fragm3: This condition needs to be changed, check is process is present first

src/utils/urls.js
```diff
@@ -1,5 +1,12 @@
...
+    return 'http://' + window.location.hostname + ':4000';
+  }
+  return 'https://api.susi.ai';
+}
+
 const urls = {
-  API_URL: process.env.REACT_APP_API_URL || 'https://api.susi.ai',
+  API_URL: check(),
   SOUND_SERVER_API_URL: 'http://0.0.0.0:7070',
```
norbusan: Why not use the same `'http://' + window.location.hostname + ':7070'; here?`

src/utils/urls.js
alok760: Writing 0.0.0.0:4000 doesn't work as its not relative and the JS tries to connect to a server running on the computer in which its running

fragm3:
@alok760 surge link not working

alok760:
@fragm3 updated the link

src/utils/urls.js
alok760: @norbusan do you mean `SOUND_SERVER_API_URL`, that doesn't need to be present at all as all endpoints on the control server is relative to this page,I will make an issue for it.

mariobehling: Which issue did you open?

src/utils/urls.js
alok760: @fragm3 it works fine currently, if i build without setting the `REACT_APP_LOCAL_ENV` environment variable, it builds the normal version of the page.

alok760: what do you suggest ?

src/utils/urls.js
alok760: @mariobehling i opened it just now - https://github.com/fossasia/susi.ai/issues/2806

src/utils/urls.js
fragm3: `process && process.env.REACT_APP_LOCAL_ENV === 'true'`

PR #2833, opened 2019-08-12
Admin: Provide more options to edit users

### LINKED ISSUE #2554: Admin: Provide more options to edit users
Admins are currently not able to edit the user data itself. Please provide options to edit email addresses and user names by accessing the edit button and allow admins to access the settings account of users.

Please implement:
* Change "Edit" link to "Change Role"
* Use "Edit" to direct to the setting of the user. The admin should be able to access the full user settings.

harsh-jindal:
Fixes #2554 

Changes: Clicking on edit in users tab in admin redirects to settings page of user.
Demo Link: https://pr-2833-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/components/Settings/AccountTab.react.js
+++ b/src/components/Settings/AccountTab.react.js
@@ -19,7 +19,7 @@ import appActions from '../../redux/actions/app';
 import { voiceList } from '../../constants/SettingsVoiceConstants.js';
 import { getUserAvatarLink } from '../../utils/getAvatarProps';
 import styled from 'styled-components';
-import { setUserSettings, uploadAvatar } from '../../apis';
+import { setUserSettings, uploadAvatar, deleteUserAccount } from '../../apis';
 import defaultAvatar from '../../images/defaultAvatar.png';
 import CircularProgress from '@material-ui/core/CircularProgress';
 import isUserName from '../../utils/isUserName';
@@ -89,7 +89,13 @@ const styles = {
 class AccountTab extends React.Component {
   constructor(props) {
     super(props);
-    const { timeZone, prefLanguage, userName, avatarType } = this.props;
+    const {
+      timeZone,
+      prefLanguage,
+      userName,
+      avatarType,
+      userEmailId,
+    } = this.props;
     this.state = {
       timeZone,
       prefLanguage,
@@ -98,13 +104,14 @@ class AccountTab extends React.Component {
       avatarType,
       avatarSrc: defaultAvatar,
       file: '',
-      imagePreviewUrl: getUserAvatarLink(),
+      imagePreviewUrl: getUserAvatarLink({ userEmailId }),
       isAvatarAdded: false,
       uploadingAvatar: false,
       isAvatarUploaded: false,
       settingSave: false,
       avatarAnchorEl: null,
       loading: false,
+      deleteUserByAdminEmail: '',
     };
     if ('speechSynthesis' in window) {
       this.TTSBrowserSupport = true;
@@ -185,10 +192,11 @@ class AccountTab extends React.Component {
 
   handleAvatarSubmit = () => {
     const { file } = this.state;
-    const { accessToken, actions } = this.props;
+    const { accessToken, actions, userEmailId } = this.props;
     // eslint-disable-next-line no-undef
     let form = new FormData();
     form.append('access_token', accessToken);
+    userEmailId && form.append('email', userEmailId);
     form.append('image', file);
     this.setState({ uploadingAvatar: true });
     uploadAvatar(form).then(() => {
@@ -242,8 +250,14 @@ class AccountTab extends React.Component {
 
   handleSubmit = () => {
     const { timeZone, prefLanguage, userName, avatarType } = this.state;
-    const { actions } = this.props;
-    const payload = { timeZone, prefLanguage, userName, avatarType };
+    const { actions, userEmailId } = this.props;
+    let payload = {
+      timeZone,
+      prefLanguage,
+      userName,
+      avatarType,
+    };
+    payload = userEmailId !== '' ? { ...payload, email: userEmailId } : payload;
     this.setState({ loading: true });
     setUserSettings(payload)
       .then(data => {
@@ -256,7 +270,7 @@ class AccountTab extends React.Component {
             .then(() => {
               this.setState({ settingSave: true, loading: false });
             })
-            .then(() => actions.updateUserAvatar());
+            .then(() => userEmailId === '' && actions.updateUserAvatar());
         } else {
           actions.openSnackBar({
             snackBarMessage: 'Failed to save Settings',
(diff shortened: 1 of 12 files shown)


harsh-jindal:
Dependent on the PR https://github.com/fossasia/susi_server/pull/1383.Once that is merged the admin will be able to change the profile pic of the user. Rest is done.

harsh-jindal:
@akshatnitd Then we will have to handle two props of everything one for the current admin and one for the user. As we have one page to view the information, the admin can at once view his settings or other user's settings. So do we still need to maintain two stores?

akshatnitd:
@harsh-jindal Suppose, there is a setting for Press Enter to Send/Theme. If we use, same store, then the other user's settings would be applied to client as well. We can have same reducers and actions. We can check if an email is passed and store the values in a separate object. And the same approach, while consuming from mapStateToProps

src/components/Settings/AccountTab.react.js
```diff
@@ -454,6 +520,7 @@ function mapStateToProps(store) {
     email: store.app.email,
     accessToken: store.app.accessToken,
     avatarType: store.settings.avatarType,
+    userSettingsViewedByAdmin: store.settings.userSettingsViewedByAdmin,
```
akshatnitd: Instead of checking if email !== null everywhere, check it here and sent appropriate store values via props.

src/components/Settings/ChatAppTab.react.js
```diff
@@ -91,11 +109,13 @@ class ChatAppTab extends React.Component {
 ChatAppTab.propTypes = {
   enterAsSend: PropTypes.bool,
   actions: PropTypes.object,
+  userSettingsViewedByAdmin: PropTypes.object,
 };
 
 function mapStateToProps(store) {
   return {
     enterAsSend: store.settings.enterAsSend,
+    userSettingsViewedByAdmin: store.settings.userSettingsViewedByAdmin,
```
akshatnitd: Same here

src/components/Settings/MicrophoneTab.react.js
```diff
@@ -109,11 +122,13 @@ class MicrophoneTab extends React.Component {
 MicrophoneTab.propTypes = {
   micInput: PropTypes.bool,
   actions: PropTypes.object,
+  userSettingsViewedByAdmin: PropTypes.object,
 };
 
 function mapStateToProps(store) {
   return {
     micInput: store.settings.micInput,
+    userSettingsViewedByAdmin: store.settings.userSettingsViewedByAdmin,
```
akshatnitd: Same here

src/components/Settings/MobileTab.react.js
```diff
@@ -230,6 +262,7 @@ function mapStateToProps(store) {
     phoneNo: store.settings.phoneNo,
     countryCode: store.settings.countryCode,
     countryDialCode: store.settings.countryDialCode,
+    userSettingsViewedByAdmin: store.settings.userSettingsViewedByAdmin,
```
akshatnitd: Same here

src/components/Settings/SpeechTab.react.js
```diff
@@ -197,6 +244,7 @@ function mapStateToProps(store) {
     ttsLanguage: store.settings.ttsLanguage,
     speechPitch: store.settings.speechPitch,
     speechRate: store.settings.speechRate,
+    userSettingsViewedByAdmin: store.settings.userSettingsViewedByAdmin,
```
akshatnitd: Same here

harsh-jindal:
@akshatnitd Made the changes. Please review.

harsh-jindal:
@fragm3 The issue says to implement it in this manner.

fragm3:
Ahh, okay

src/components/Settings/ChatAppTab.react.js
```diff
@@ -28,10 +29,14 @@ class ChatAppTab extends React.Component {
...
   handleSubmit = () => {
-    const { actions } = this.props;
+    const { actions, userEmailId } = this.props;
     const { enterAsSend } = this.state;
     this.setState({ loading: true });
-    setUserSettings({ enterAsSend })
+    let payload = {
+      enterAsSend,
+    };
+    payload = userEmailId !== '' ? { ...payload, email: userEmailId } : payload;
```
akshatnitd: I think we can always send the email in payload. We can have a check while setting in store, if the email is equal to logged in user email, then use the store that already existed, else use `userSettingsViewedByAdmin` object.

harsh-jindal:
@akshatnitd Made the changes. It turns out that was never required because if the the user is accessing the settings page the userEmailId will always be empty.

src/components/Settings/AccountTab.react.js
```diff
@@ -444,16 +492,26 @@ AccountTab.propTypes = {
...
 
 function mapStateToProps(store) {
+  const userSettingsViewedByAdmin = store.settings.userSettingsViewedByAdmin;
+  const {
+    email,
+    userName,
+    timeZone,
+    prefLanguage,
+    avatarType,
+  } = userSettingsViewedByAdmin;
```
akshatnitd: Can we write, something like.

`const settings = email !== '' ? store.settings.userSettingsViewedByAdmin : store.settings;`

Then userName can be `settings.userName`

harsh-jindal:
@akshatnitd Made the changes.
Also I need to put a check for payload, and cannot send email always in the call because on the server if the user is an admin and is we send `email:userEmailId` always then his settings will not be saved as email will be an empty string.
On the server condition is `(userRole == admin or userRole==superAdmin) and email != null` as `empty string` is `!=null` we end up searching for a user with email as empty string.

src/redux/reducers/settings.js
```diff
@@ -100,11 +136,28 @@ export default handleActions(
           button: themeArray[5],
         },
       };
+      let userSettingsViewedByAdmin = email
+        ? { ...userSettings, email }
+        : defaultState.userSettingsViewedByAdmin;
+      userSettings = email ? { ...state } : userSettings;
```
fragm3: Fix this condition, you are adding state twice if email exists.

src/redux/reducers/settings.js
```diff
@@ -33,12 +33,48 @@ const defaultState = {
   messageBackgroundImage: '',
   avatarType: 'default',
   devices: {},
+  userSettingsViewedByAdmin: {
```
fragm3: Instead of initialising with same values, create a `constant config` in the same file and use it for default state, userSettingsViewedByAdmin

src/redux/reducers/settings.js
```diff
@@ -100,11 +136,28 @@ export default handleActions(
...
+      };
     },
     [actionTypes.SETTINGS_SET_USER_SETTINGS](state, { payload }) {
+      let userSettings = { ...payload };
+      const { email } = state.userSettingsViewedByAdmin;
+      let userSettingsViewedByAdmin =
+        email !== ''
+          ? { ...state.userSettingsViewedByAdmin, ...userSettings }
+          : defaultState.userSettingsViewedByAdmin;
+      userSettings = email !== '' ? { ...state } : userSettings;
```
fragm3: Same here

harsh-jindal:
@fragm3 Made the changes.

PR #2853, opened 2019-08-14
Hide sort option when no skills are present

### LINKED ISSUE #2825: Hide Sort option when no skills are present
**Actual Behaviour**

Capture

**Expected Behaviour**

When no skills are present, hide sort & filter option.

**Steps to reproduce it**

**LogCat for the issue**

**Screenshots of the issue**

Yes, I would like to work on this.

plxity:
Fixes #2825 

Changes: Sort option is not visible when no skills are present.

Demo Link: https://pr-2853-fossasia-susi-web-chat.surge.sh

Capture


harsh-jindal:
I think even if we want to remove the sort options the message should be aligned to the centre rather than on the left.

plxity:
@fragm3 I think we should not display something which is not getting used at that point.

plxity:
@akshatnitd Okay, I'll make the changes.

PR #2856, opened 2019-08-15
Fix menu under ? is not completely inside the screen

### LINKED ISSUE #2671: menu under "?" is not completely inside the screen
**Actual Behaviour**
menu that pops by hovering on "?" is not completely inside the screen 
susi1
**Expected Behaviour**
menu should be completely inside the the screen

Yes , i would like to work on it

harsh-jindal:
Fixes #2671 

Changes: `?` menu is completely inside the screen.
Demo Link: https://pr-2856-fossasia-susi-web-chat.surge.sh

Screenshot 2019-08-15 at 4 44 35 PM

### DIFF
--- a/src/components/NavigationBar/SearchBar.js
+++ b/src/components/NavigationBar/SearchBar.js
@@ -26,8 +26,20 @@ const StyledSearchBar = styled(_SearchBar)`
   height: 35px;
   padding-left: ${props => `${props.paddingLeft}rem`};
   width: 39rem;
-  @media (max-width: 1280px) {
+  @media (max-width: 1380px) {
     width: 30rem;
+  }
+  @media (max-width: 1250px) {
+    width: 25rem;
+  }
+  @media (max-width: 1150px) {
+    width: 21rem;
+  }
+  @media (max-width: 1100px) {
+    width: 19rem;
+  }
+  @media (max-width: 1050px) {
+    width: 14.5rem;
   }
   @media (max-width: 960px) {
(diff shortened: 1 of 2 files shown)


harsh-jindal:
@fragm3 It is fixed for view below 1250px. You need to reload the page when you change dimensions to see the fix.

fragm3:
@harsh-jindal we definitely need email/name

harsh-jindal:
@fragm3 We don't have it for mobile view, so I think its fine to remove email and keep search bar big. Or we need to reduce the size of the search bar significantly, which looks odd for screens around 1000-1100px. There are other two ways:
1. keep the username and decrease the width of the search bar for views around 1100px.
2. keep the username and keep only search icon.
Which one should i go with ?

fragm3:
@harsh-jindal decrease search bar width for that view port

harsh-jindal:
@fragm3 Fixed.

harsh-jindal:
@samagragupta As it as been already said before we are not aiming for such small view ports.

harsh-jindal:
@fragm3 Fixed.

harsh-jindal:
@akshatnitd Then I suggest we use the `+` icon instead of  `create` text below `400px` because we cannot fit the current navbar in 360px. Shall I change it to that ?

harsh-jindal:
@akshatnitd Made the changes. 
Screenshot 2019-08-17 at 6 39 14 PM

PR #2905, opened 2019-09-10
made changes to scroll in skillcard

### LINKED ISSUE #2900: Scroll button appears when content does not overflow
**Actual Behaviour**

Scroll button appears even when content in it does not overflow.

**Expected Behaviour**

Scroll button should not appear in that case.

**Screenshots of the issue**

Screenshot from 2019-09-04 18-35-1121

Yes

Dilshaad21:
Fixes #2900 

Changes: Scroll button appears when the scroll cards exceed maxwidth of the window

Demo Link: https://pr-2905-fossasia-susi-web-chat.surge.sh

Screenshots of the change: 
**Before**
Screenshot from 2019-09-04 18-35-1121

**After**
new

### DIFF
--- a/src/components/cms/SkillCardScrollList/SkillCard.js
+++ b/src/components/cms/SkillCardScrollList/SkillCard.js
@@ -240,7 +240,7 @@ class SkillCard extends Component {
         );
       },
     );
-    if (cards.length <= this.state.scrollCards) {
+    if (cards.length <= this.state.scrollCards + 1) {
       this.setState({ rightBtnDisplay: 'none', leftBtnDisplay: 'none' });
     }
     this.setState({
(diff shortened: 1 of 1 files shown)


Dilshaad21:
@samagragupta updated demo link

Dilshaad21:
@akshatnitd I suppose that the state is not getting updated by updatewindowdimensions()

Dilshaad21:
@akshatnitd used it in `componentWillMount()` so that it can update the state before `loadskillcards`
. Is it OK?

src/components/cms/SkillCardScrollList/SkillCard.js
```diff
@@ -62,14 +62,19 @@ class SkillCard extends Component {
     };
   }
 
+  componentWillMount = () => {
```
akshatnitd: The use of this React lifecycle is now deprecated. Please try using some other approach.

Dilshaad21: I hope this approach works fine.

PR #2911, opened 2019-09-17
Resolved issue of Doublication of messages #2851

### LINKED ISSUE #2851: Doublication of messages
**Actual Behaviour**
When you ask "who is Mario" the answer is
2 times "I’d rather not say."
or
2 Times Why do you ask?

**Expected Behaviour**
Only once.

Vineet-Sharma29:
Fixes #2851 

Changes: Following changes have been made:-
1. in src/utils/generateMessage.js in line 39 forEach loop has been removed because in some cases multiple actions were obtained due to which susiReply were multiple
2. it is replaced by taking only the first entry from the actions array

Demo Link: https://pr-2911-fossasia-susi-web-chat.surge.sh

Screenshots of the change: 

**Before**

Screenshot from 2019-09-17 20-00-58

**After**

Screenshot from 2019-09-17 20-04-03

### DIFF
--- a/src/utils/generateMessage.js
+++ b/src/utils/generateMessage.js
@@ -36,6 +36,12 @@ export default function({
               response,
               voice,
             }).then(susiMessages => {
+              susiMessages = Array.from(
+                new Set(susiMessages.map(message => message.text)),
+              ).map(text => {
+                return susiMessages.find(message => message.text === text);
+              });
+
               if (susiMessages[0].planDelay) {
                 tasks.push(
                   plannedFunction,
--- a/src/utils/generateMessage.js
+++ b/src/utils/generateMessage.js
@@ -36,17 +36,16 @@ export default function({
               response,
               voice,
             }).then(susiMessages => {
-              susiMessages.forEach(eachMessage => {
-                if (eachMessage.planDelay) {
-                  tasks.push(
-                    plannedFunction,
-                    () => createSusiMessage({ message: eachMessage }),
-                    eachMessage.planDelay,
-                  );
-                } else {
-                  createSusiMessage({ message: eachMessage });
-                }
-              });
+              if (susiMessages[0].planDelay) {
+                tasks.push(
+                  plannedFunction,
+                  () => createSusiMessage({ message: susiMessages[0] }),
+                  susiMessages[0].planDelay,
+                );
+              } else {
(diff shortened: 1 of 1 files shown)


src/utils/generateMessage.js
```diff
@@ -36,17 +36,16 @@ export default function({
...
-                  tasks.push(
-                    plannedFunction,
-                    () => createSusiMessage({ message: eachMessage }),
-                    eachMessage.planDelay,
-                  );
-                } else {
-                  createSusiMessage({ message: eachMessage });
-                }
-              });
+              if (susiMessages[0].planDelay) {
```
akshatnitd: This is the incorrect fix, we need all the messages that comes in the `susiMessages` array. This piece of code was added to make Planned actions work.

Vineet-Sharma29: ok @akshatnitd I will fix it removing duplicate entries from the `susiMessages` array and then runnnig `forEach` loop over the array so as to print all the messages. Can you suggest if this approach is correct?

akshatnitd: Does `susiMessages` contain duplicate entries? Can you give an example case?

Vineet-Sharma29: @akshatnitd when I have query : `who is mario` , I get following list of actions(as observed in networking tab of console):-
`
"actions": [
      {
        "language": "en",
        "type": "answer",
        "expression": "Great question. I never actually thought about that before, and I am not sure."
      },
      {
        "type": "answer",
        "mood": "sabta",
        "expression": "Great question. I never actually thought about that before, and I am not sure."
      }
    ],
`

Vineet-Sharma29: @akshatnitd changes has been done so as to include non-duplicate, multiple messages

akshatnitd: @Vineet-Sharma20 For this specifically, this should be fixed from backend and only 1 action should be ideally coming from backend.

Vineet-Sharma29: @akshatnitd ok

Vineet-Sharma29: @akshatnitd I have opened this as an issue in susi_server

akshatnitd: @Vineet-Sharma29 Thanks! I think you can close this PR as it is a server issue!

PR #2927, opened 2019-10-01
Set "default" avatar selected by default

### LINKED ISSUE #2919: Default avatar not selected in settings
**Actual Behaviour**

Default avatar is set as avatar but not showing selected in settings/user

**Expected Behaviour**

By default "default" avatar should be shown as current avatar and selected as well. 

**Steps to reproduce it**
1.Go to settings
2.Go to account

**Screenshots of the issue**

**Actual**
im2

**Expected**
im1

Yes

Dilshaad21:
Fixes #2919 

Changes: Set the default avatar as "default" when initially no avatar is selected by a new user.

Demo Link: https://pr-2927-fossasia-susi-web-chat.surge.sh

Screenshots of the change: 
**Before**
im2

**After**
im1

### DIFF
--- a/src/redux/reducers/settings.js
+++ b/src/redux/reducers/settings.js
@@ -75,7 +75,7 @@ export default handleActions(
         checked = config.checked,
         backgroundImage = config.backgroundImage,
         messageBackgroundImage = config.messageBackgroundImage,
-        avatarType = config.messageBackgroundImage,
+        avatarType = config.avatarType,
       } = settings;
       let { customThemeValue } = settings;
       const themeArray = customThemeValue


src/components/Settings/AccountTab.react.js
```diff
@@ -372,7 +372,7 @@ class AccountTab extends React.Component {
             <div>
               <Select
                 onChange={this.handleAvatarTypeChange}
-                value={avatarType}
+                value={avatarType ? avatarType : 'default'}
```
akshatnitd: Try to change the default avatarType value in the reducer level, so that it can reflect here.

Dilshaad21:
@akshatnitd done changes in the reducer level.

PR #2932, opened 2019-10-03
updated the password notification

### LINKED ISSUE #2704: Unnecessary notification below password in the Login Panel
**Actual Behaviour**
In the Login panel, below password, it show the error notification -"Password should be at least 6 to 64 characters",which is totally unnecessary because user have to input the exact password that he/she used during signup.

**Expected Behaviour**

The following password error notification should be removed 

**Screenshots of the issue**
pass_Comm

Yes ,i would like to work on it

tanujvyas10:
updated the unnecessary password notification

Fixes #2704 

Demo:-https://pr-2932-fossasia-susi-web-chat.surge.sh
Changes: Updated the notification

### DIFF
--- a/src/components/Auth/Login/Login.react.js
+++ b/src/components/Auth/Login/Login.react.js
@@ -160,13 +160,9 @@ class Login extends Component {
       }
       case 'password': {
         const password = event.target.value.trim();
-        let passwordErrorMessage = '';
-        if (!password || password.length < 6) {
-          passwordErrorMessage = 'Password should be atleast 6 characters';
-        }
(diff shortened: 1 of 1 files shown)


tanujvyas10:
@plxity sir, currently there is the notification in login's password section-"password should be at least 6 characters", and also the login button is enabling after the 6 or more characters are typed but we know that such notification never occurs in any login form because a user will always enter the exact password with exact length that she/he has signed in before.
susi4

akshatnitd:
This is already fixed. So closing it.

PR #2965, opened 2019-10-18
Fixes 2964 - Addition of tooltips

### LINKED ISSUE #2964: No way to identify the menu bar buttons
**Actual Behaviour**

The menubar buttons on the top right corner(dashboard and message) give no description or any hints regarding what their purpose is. A First time user would not be able to identify the use of those buttons without having to click on them.

**Expected Behaviour**

A tooltip is displayed when hovering over the button briefly describing what it does.

**Screenshots of the issue**

noData

Yes

AuraOfDivinity:
Fixes #2964 

Changes: 

1) Added tooltips to help better identify the buttons.

Screenshots of the change: 

AdditionOfToolltips


src/components/NavigationBar/index.js
```diff
@@ -633,13 +634,28 @@ class NavigationBar extends Component {
...
                       </div>
                     </StyledIconButton>
-                    <IconButton
-                      color="inherit"
-                      onClick={() => history.push('/dashboard')}
-                      style={{ padding: '7px' }}
+                    <Tooltip
+                      placement="bottom"
+                      title="Dashboard"
+                      PopperProps={{
```
akshatnitd: Code is repetitive here. Please make a common component, so that we can use Tooltip in the following pattern.
```
<Tooltip text="Dashboard">
   <Children />
</Tooltip>

```

PR #2982, opened 2019-10-23
Fixes 2980 - Fixing issues with the empty username field in Account tab

### LINKED ISSUE #2980: Cannot set timezone or default language.
**Actual Behaviour**

Currently in settings -> Account it's unable to set a language or a timezone without having set up a username first . If a user tries to update the timezone or default language with an empty username field, it returns a 400 Bad Request error. 

**Expected Behaviour**

If the username field is kept empty, the user should be prompted to fill it in before sending the server request.

**Screenshots of the issue**

CannotChangeTimezone

Yes

AuraOfDivinity:
Fixes #2980

Changes: 
Added additional validation so that the request is only sent after the username field is filled in(when a valid username is not set).

Demo Link :http://mighty-slip.surge.sh/

Screenshots of the change:
SavingAccountSettings

### DIFF
--- a/src/components/Settings/AccountTab.react.js
+++ b/src/components/Settings/AccountTab.react.js
@@ -247,9 +247,24 @@ class AccountTab extends React.Component {
     });
   };
 
+  isUserNameEntered = () => {
+    if (document.getElementById('userNameField').value.length === 0) {
+      this.setState({
+        userNameError:
+          'A valid username is required before setting timezone or preferred language!',
+      });
+      return false;
+    }
+    return true;
+  };
+
   handleSubmit = async () => {
     const { timeZone, prefLanguage, userName, avatarType } = this.state;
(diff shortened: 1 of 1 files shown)


src/components/Settings/AccountTab.react.js
```diff
@@ -247,9 +247,24 @@ class AccountTab extends React.Component {
...
+      return false;
+    }
+    return true;
+  };
+
   handleSubmit = async () => {
     const { timeZone, prefLanguage, userName, avatarType } = this.state;
     const { actions, userEmailId } = this.props;
+
+    if (!this.isUserNameEntered()) {
```
akshatnitd: Checking `userName.length > 0` would be sufficient to check the condition.

gorfadvijay:
close this issue already solved pr 3180

PR #2984, opened 2019-10-24
Only set loadingdrafts to false when there are no drafts to display

### LINKED ISSUE #2983: Catches error when no drafts created
**Actual Behaviour**

The snackBar `Couldn't get your drafts. Please reload the page.` pops up as the catch function runs while loading drafts. But it should not pop up when there are no created.

**Expected Behaviour**

The `snackBar` must not appear in that case.

**Steps to reproduce it**

1. Login to your account
2. Go to my bots.

**Screenshots of the issue**

fff

Yes, I would like to work on this issue.

Dilshaad21:
Fixes #2983 

Changes: 
-Added an `else` statement in `showDrafts()` to only set `loadingDrafts` to false when there are no drafts. 
-In `getDrafts()` it jumps to `catch()` only when error occurs while receiving the `promise` in `let payload = await readDraft()`.

Demo Link: https://pr-2984-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/components/cms/BotBuilder/BotBuilder.js
+++ b/src/components/cms/BotBuilder/BotBuilder.js
@@ -287,10 +287,12 @@ class BotBuilder extends React.Component {
       const { drafts } = payload;
       this.showDrafts(drafts);
     } catch (error) {
-      actions.openSnackBar({
-        snackBarMessage: "Couldn't get your drafts. Please reload the page.",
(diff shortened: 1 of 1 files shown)


Dilshaad21:
@akshatnitd I think this is a server related issue. Closing pr.

PR #3020, opened 2019-11-05
Center aligned no skills text

### LINKED ISSUE #3003: No result text should be center aligned
**Actual Behaviour**

No result text is left aligned.
**Expected Behaviour**

No result text should be center aligned.

**Screenshots of the issue**

Screenshot 2019-10-30 at 1 25 40 AM

Yes

kpulkit29:
Fixes #3003 

Changes: Earlier the no skills text was inside the grid which had defined columns so it wasn't possible to center align the the individual grid with respect to the screen. Have placed the text in a new division (instead of reusing that in grid) which shows up in case of no skills found.

Demo Link: https://pr-3020-fossasia-susi-web-chat.surge.sh

Screenshot 2019-11-05 at 1 09 59 AM
Screenshot 2019-11-05 at 1 07 45 AM
Screenshot 2019-11-06 at 1 05 12 AM

### DIFF
--- a/src/components/cms/BrowseSkill/BrowseSkill.js
+++ b/src/components/cms/BrowseSkill/BrowseSkill.js
@@ -617,7 +617,8 @@ class BrowseSkill extends React.Component {
     let metricsHidden =
       routeType || searchQuery.length > 0 || ratingRefine || timeFilter;
 
-    let renderSkillCount = '';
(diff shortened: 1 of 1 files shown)


kpulkit29:
@akshatnitd I have added styled components to it

PR #3024, opened 2019-11-06
Close button fix

### LINKED ISSUE #2947: Allignment issue with scroll bar
**Actual Behaviour**

The vertical scroll bar comes over close button while selecting colors in custom theme.

**Expected Behaviour**

It should not come over it.

**Steps to reproduce it**

1. Go to settings
2. Go to Theme
3. Select Custom theme->Edit theme.

**Screenshots of the issue**

newr

Yes

vaibhavmathur13:
Fixes #2947 

Changes: Shifted the close button to left so that it doesn't overlap with scroll bar.

Demo Link: https://pr-3024-fossasia-susi-web-chat.surge.sh/

Screenshots of the change: Before:
Screenshot from 2019-11-06 15-13-32
After:
Screenshot from 2019-11-06 15-12-08


Dilshaad21:
@vaibhavmathur13 is there any alternate way you can shift the close button just above the scroll bar.

plxity:
@vaibhavmathur13 The suitable solution to this is to set the `overflow : hidden` for the highlighted `div` element. It would remove the unnecessary scroll bar and solve the issue.
Screenshot from 2019-11-07 13-42-41

vaibhavmathur13:
Screenshot from 2019-11-07 23-31-04
@plxity The scroll bar is not unnecessary, it is used to scroll down the color picker of last color box. Removing scroll bar will not allow the users to see the complete color picker.

vaibhavmathur13:
@Dilshaad21 I don't think we can move the close button on top of scroll bar. If you have any ideas about it , I can implement and send another PR. I've suggested some other options on gitter. Thanks.

vaibhavmathur13:
The file in which we are making changes is only for the close button.
In which file we have to make changes so that is reflects in the container.
Where should I look for the file. Any suggestions?
Keep me posted. Thanks.

vaibhavmathur13:
@akshatnitd @Dilshaad21

PR #3047, opened 2019-11-13
Implements #3039 - Improved video player

### LINKED ISSUE #3039: Videos played on the chat are too small.
**Actual Behaviour**

Currently when a video is played using play command in the chat, the resulting video component displaying the video is too small. 

**Expected Behaviour**

Display the video is a separate larger popup or enlarge the chat when a video is playing(if the screen size permits it).

**Screenshots of the issue**

image

Yes

AuraOfDivinity:
Fixes #3039 

Changes: 
1) When a youtube video is requested using the play command, a popout button will now appear underneath the video. If the user wants he/she can click on this popout button to view the video on a much larger popout.
2) This popout button is only available in larger screen sizes and is not available on smaller devices.

Demo Link: https://pr-3047-fossasia-susi-web-chat.surge.sh

Screenshots:

ImprovedVideo

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "react-modal": "3.8.1",
     "react-orgchart": "1.0.5",
     "react-piwik": "1.6.0",
+    "react-player": "^1.14.1",
     "react-ratings-declarative": "3.4.1",
     "react-recaptcha": "2.3.10",
     "react-redux": "7.0.3",
--- a/src/components/ChatApp/MessageListItem/MessageListItem.react.js
+++ b/src/components/ChatApp/MessageListItem/MessageListItem.react.js
@@ -28,6 +28,7 @@ class MessageListItem extends React.Component {
       play: false,
       width: this.props.showChatPreview ? 234 : 384,
       height: this.props.showChatPreview ? 168 : 240,
+      showModal: false,
     };
   }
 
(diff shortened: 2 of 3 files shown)


AuraOfDivinity:
@akshatnitd Thanks for the feedback! The Icon you suggested looks similar to the youtube's full screen icon(which could be seen on the bottom right when the video starts playing.) Wouldn't it be confusing to the user to 2 similar icons?

AuraOfDivinity:
@akshatnitd I gave it a thought on adding the icon you suggested _on top of the player_. However I still feel like the player is cluttered with so many different information/icons that adding our icon there would only make it look even worse. Please check the image on the bottom. (Note the icons, video title, channel icon etc on the player). I still feel like having the icon on the bottom left would be the most suitable place. Would like to hear your thoughts on this.

I also searched different icon libraries including the material ui icon library, fa icon library etc but couldn't really find an exact copy of the icon you suggested. The closest one I could find was the one used in the following image. (https://material.io/resources/icons/?search=Aspe&icon=aspect_ratio&style=baseline)

video2

In the last commit I changed the button to an icon and also added in a tooltip. Please check the deployment and let me know your thought! https://pr-3047-fossasia-susi-web-chat.surge.sh

PR #3129, opened 2019-12-22
Fixes 2980 - Fixing issues with the empty username field in Account tab

### LINKED ISSUE #2980: Cannot set timezone or default language.

gorfadvijay:
Fixes #2980 - Fixing issues with the empty username field in Account tabcreenshots of the change: 

Demo Link: https://pr-3129-fossasia-susi-web-chat.surge.sh

Screenshots of the change:
fixed2980

### DIFF
--- a/src/components/Settings/AccountTab.react.js
+++ b/src/components/Settings/AccountTab.react.js
@@ -250,6 +250,14 @@ class AccountTab extends React.Component {
   handleSubmit = async () => {
     const { timeZone, prefLanguage, userName, avatarType } = this.state;
     const { actions, userEmailId } = this.props;
+    if (!(userName.length > 0)) {
+      this.setState({
+        userNameError:
+          'A valid username is required before setting timezone or preferred language!',
+      });
+      return;
+    }
+
     let payload = {
       timeZone,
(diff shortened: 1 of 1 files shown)


src/components/Settings/AccountTab.react.js
```diff
@@ -250,6 +250,14 @@ class AccountTab extends React.Component {
   handleSubmit = async () => {
     const { timeZone, prefLanguage, userName, avatarType } = this.state;
     const { actions, userEmailId } = this.props;
+    if (!(userName.length > 0)) {
```
akshatnitd: Use `userName.trim().length` to remove blank spaces

gorfadvijay: changed it !

PR #3134, opened 2019-12-23
Add Service Workers

### LINKED ISSUE #3133: Service workers needed
**Actual Behaviour**
There are no _service workers_. [Service workers](https://developers.google.com/web/fundamentals/primers/service-workers) increase the performance of app. It loads the content instantly, when user visits the app more than once

**Expected Behaviour**
Service workers should be added to the app

Yes

MuhammadWasif:
Fixes #3133 
### Changes: 
Service workers have been added

### Demo Link: 
https://sad-visvesvaraya-f4a745.netlify.com/

Screenshots of the change: 
There are no changes in UI so there's no need of screenshot


MuhammadWasif:
> The PR looks good. But, how will be ensure the cache version to update? Do we need to change the name of cache explicitly on every deployment?

`<meta http-equiv="Cache-Control" content="max-age=2628002">`
Adding this in head tag will automatically update caches after 1 month. Let me add this 🚀🚀

public/index.html
```diff
@@ -21,6 +22,8 @@
     <title></title>
     <meta name="keywords" content="Personal Assistant Free Software Artificial Intelligence Chat Bot Machine Learning Scientific User Support IoT"/>
     <meta name="description" content="Chat with an Open Source Personal Assistant"/>
+    <meta name="theme-color" content="#4285f4">
+    <meta http-equiv="Cache-Control" content="max-age=2628002">
```
akshatnitd: This approach would let the cache expire only after a month, and we have deployments made almost daily. So, the latest JS bundle wouldn't be installed as, it would still use it from cache. We need to add some config which changes the cache name on every deployment.

MuhammadWasif: We can make it update after a day. But I think there can't be some approach to change cache name on every deployment.

MuhammadWasif: But if we cache only ``build/static/css/**.css`` and ``build/static/js/**.js`` then it will only cache styles and JavaScript instead of whole app. Can I do this?

akshatnitd: @MuhammadWasif The bundled files also goes into the static folder.

MuhammadWasif: So, can we make it update after a day? I think that will be a good approach.

MuhammadWasif:
Closing because of no response!

Issue #3175, opened 2019-12-31
Substitute external dependencies - Avoid calling CDNs, external JS, Fonts etc.


mariobehling:
The web client should be ready for use in a local network without Internet access. This is also important for privacy. People don't always want to call Google fonts and share their IP address and other data automatically with external services (by calling these dependencies). 

Please substitute external dependencies like Google fonts, external scripts etc. and ensure all required components are part of the repo and deployed site itself.

anishagg17:
@mariobehling  I am working on this issue

PR #3189, opened 2020-01-02
removed slick dependency

anishagg17:
"First steps for issue" #3175

Changes: Removed slick dependency . It was about 80 lines of code for which a request to an external server was made

Demo Link: https://pr-3189-fossasia-susi-web-chat.surge.sh


iamareebjamal:
Even this should be reverted

PR #3201, opened 2020-01-03
Added loader for skillShow

### LINKED ISSUE #3200: No loader for SkillSlideshow
**Actual Behaviour**

On HomeScreen SkillSlideshow takes time to load and hence it appears differently 

**Expected Behaviour**

there should be loader for SkillSlideshow

**Screenshots of the issue**

Screenshot 2020-01-03 at 10 22 56 PM

Yes

anishagg17:
Fixes #3200

Changes: Added Loader for skillShow component 

Demo Link: https://pr-3201-fossasia-susi-web-chat.surge.sh

Screenshots of the change: 
Screenshot 2020-01-03 at 10 25 07 PM


anishagg17:
Mobile View:

Screenshot 2020-01-04 at 6 49 23 PM

Desktop view:

Screenshot 2020-01-04 at 6 49 46 PM

src/components/cms/BrowseSkill/skillLoader.js
```diff
@@ -34,11 +34,43 @@ const addLoader = () => {
 
 const SkillLoader = () => {
   return (
```
akshatnitd: Call isMobileView only once here and store it in a variable. It will minimize the function calls. Also, can you fix the card loader in the mobile view?

src/components/cms/BrowseSkill/skillLoader.js
```diff
@@ -7,6 +7,8 @@ const Container = styled.div`
   display: flex;
 `;
 
+const isMobile = isMobileView();
```
anishagg17: sir, I have added it here
so that this variable is available to both functions instead of just one function

src/components/cms/BrowseSkill/skillLoader.js
anishagg17: > Call isMobileView only once here and store it in a variable. It will minimize the function calls.

Done sir!

src/components/cms/BrowseSkill/skillLoader.js
akshatnitd: @anishagg17 Adding it here, won't update the values on re-render.

src/components/cms/BrowseSkill/skillLoader.js
anishagg17: > Also, can you fix the card loader in the mobile view?

Yes I have noticed that in mobile view those cards are not as good as they need to be
I have opened an issue #3209 related to it and soon will fix it

anishagg17:
> @anishagg17 Adding it here, won't update the values on re-render.

Done Sir!
But I don't think it needs to be  re-render as screen size won't change

anishagg17:
> The size will change in mobile when switching from Portrait to Landscape layout.

Okay Nice catch

PR #3228, opened 2020-01-06
First page of search result

### LINKED ISSUE #3087: Not landing into 1st page of search results
**Actual Behaviour**

The pages remain the same even if the search results are varied.

**Expected Behaviour**

On varying the search results, it should land on page 1 of search results.

**Steps to reproduce it**
Go to particular rating results and change the page, after that try to change the rating and check out the result page. 

**Screenshots of the issue**

Screencast-from-Saturday-23-November-2019-12_00_38-IST

Yes

simranaggarwal1999:
Fixes #3087 

Changes: Fixed not landing into 1st page of search results

Demo Link: https://pr-3228-fossasia-susi-web-chat.surge.sh

Screenshots of the change:

incorrect landing

correct landing

### DIFF
--- a/src/components/cms/SkillCardScrollList/SkillCard.js
+++ b/src/components/cms/SkillCardScrollList/SkillCard.js
@@ -20,7 +20,7 @@ import {
   Image,
   TitleContainer,
   RatingContainer,
-  TotalRating,
+  AverageRating,
 } from '../SkillCardStyle';
 
 const ScrollWrapper = styled.div`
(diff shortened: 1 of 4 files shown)


src/components/cms/SkillCardScrollList/SkillCard.js
```diff
@@ -20,7 +20,7 @@ import {
   Image,
   TitleContainer,
   RatingContainer,
-  TotalRating,
+  AverageRating,
```
anishagg17: Please make changes related to the issue only

simranaggarwal1999: @anishagg17 I will correct it but now the site is not running on the localhost when I reforked the repository.
image
How should I solve this.

anishagg17: Ya the server is not working well since 2 days thats why you are facing these errors because of cors parameter

It 's not because you reforked the repository.

simranaggarwal1999: @anishagg17 Ohh Okay ! When the error would get rectified?

PR #3230, opened 2020-01-07
Show table in Analytics only when content is available

### LINKED ISSUE #3229: Table should not get displayed when there is no content to show
**Actual Behaviour**

Table in the dashboard gets displayed even when there is no content to show in `Ratings` and `Skills`.

**Expected Behaviour**

It should not get displayed in that case.

Yes

Dilshaad21:
Fixes #3229  

Changes: 
- Set proper conditions in `Analytics`, `Skills` and `Ratings` section to display `Material Table` only when some content is available to show.

Demo Link: https://pr-3230-fossasia-susi-web-chat.surge.sh

Screenshots of the change:
**Before Changes**
scsdv
 
**After Changes**
csdc

### DIFF
--- a/src/components/cms/Dashboard/MyRatings.js
+++ b/src/components/cms/Dashboard/MyRatings.js
@@ -77,76 +77,80 @@ class MyRatings extends Component {
           <TableSleleton />
         ) : (
           <TableWrap>
-            <MaterialTable
-              title="My Rating"
-              columns={[
-                {
-                  title: 'Skill Name',
-                  field: 'skillName',
-                  render: rowData => {
-                    return (
-                      <Link
-                        to={{
-                          pathname: `/${rowData.group}/${getSkillFromRating(
-                            rowData.skillName,
-                          )}/${rowData.language}`,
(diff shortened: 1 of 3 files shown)


src/components/cms/Dashboard/MyRatings.js
```diff
@@ -77,76 +77,78 @@ class MyRatings extends Component {
...
-                          )}/${rowData.language}`,
-                        }}
-                      >
-                        {(
-                          rowData.skillName.charAt(0).toUpperCase() +
-                          rowData.skillName.slice(1)
-                        ).replace(/[_-]/g, ' ')}
-                      </Link>
-                    );
+            {ratingsData.length !== 0 && (
```
akshatnitd: Please add a  safety check here `ratingsData && Array.isArray(ratingsData) && ratingsData.length`

Dilshaad21: Ok then, doing the same changes to `Skills` and `Analytics` as well

PR #3246, opened 2020-01-10
Missing checks fixed that cause website to break

### LINKED ISSUE #3245: API response should not leave the website blank
**Actual Behaviour**

The website goes blank due to unexpected response from the API . Some missing checks cause it to go blank.  At some places the code is directly accessing object properties which can always be error prone and may cause the following code to break. Currently the app.js file in reducer and SkillCardScrollList.js throw error  whenever the API is down.

**Expected Behaviour**

Checks should be there instead of directly accessing the object properties

**Screenshots of the issue**

Screenshot 2020-01-10 at 4 18 13 PM
Screenshot 2020-01-10 at 4 19 43 PM

yes

kpulkit29:
Fixes #3245 

Changes: added check so that object properties are not accessed directly

Demo Link: https://pr-3246-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/components/cms/SkillCardScrollList/SkillCardScrollList.js
+++ b/src/components/cms/SkillCardScrollList/SkillCardScrollList.js
@@ -60,7 +60,7 @@ const SkillCardScrollList = ({ metricSkills, history, isMobile }) => {
   let renderCardScrollList = '';
 
   renderCardScrollList = skillCardListData.map((data, index) => {
-    return metricSkills[data.skills].length ? (
+    return metricSkills[data.skills] && metricSkills[data.skills].length ? (
       <Container key={index}>
         <HeaderText>{data.heading}</HeaderText>
         <SkillCard
--- a/src/redux/reducers/app.js
+++ b/src/redux/reducers/app.js
@@ -59,9 +59,12 @@ export default handleActions(
         return { ...state };
       }
       let keyValueObj = {};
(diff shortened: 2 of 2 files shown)


src/components/cms/SkillCardScrollList/SkillCardScrollList.js
```diff
@@ -60,7 +60,7 @@ const SkillCardScrollList = ({ metricSkills, history, isMobile }) => {
   let renderCardScrollList = '';
 
   renderCardScrollList = skillCardListData.map((data, index) => {
-    return metricSkills[data.skills].length ? (
+    return metricSkills[data.skills] && metricSkills[data.skills].length ? (
```
anishagg17: Duplicate of #3222

src/redux/reducers/app.js
```diff
@@ -59,9 +59,12 @@ export default handleActions(
         return { ...state };
       }
       let keyValueObj = {};
-      for (let [key, value] of Object.entries(keys)) {
-        keyValueObj = { ...keyValueObj, [key]: value.value };
+      if(keys) {
```
anishagg17: Duplicate of #3222

anishagg17:
please don't Duplicate #3222

akshatnitd:
Closing the PR, since it is duplicate.

PR #3258, opened 2020-01-11
Fixed minor checks

anishagg17:
Fixes #3259

Changes:

Fixed minor checks

PR #3279, opened 2020-01-14
Removed Error message for password in login window

### LINKED ISSUE #3277: Error message display while login and entering password
**Actual Behaviour**

An Error msg of "Password should be atleast 6 character" when we try to login and enter the password.

**Expected Behaviour**

While logging in, there should be no error message displayed to the user. The error message should always pop up when the password don't match

**Steps to reproduce it**

Start entering the password in the login in window.

**Screenshots of the issue**

Screenshot (81)

Yes

manan2501:
## Changes:
Removed the password length error which was shown in the login window.
Also edited src/components/About/Blog/index.js because it was showing suggestion while compiling.

### Before:
image

### After:
error_gone

### DIFF
--- a/src/components/About/Blog/index.js
+++ b/src/components/About/Blog/index.js
@@ -229,7 +229,7 @@ class Blog extends Component {
   getPosts = async offset => {
     this.setState({ postRendered: false });
     let payload = null;
-    if (this.state.nextPosts.length == 0 || offset <= this.state.startPage) {
+    if (this.state.nextPosts.length === 0 || offset <= this.state.startPage) {
       payload = await getBlogReponse(this.blogKey, 10, offset);
     } else {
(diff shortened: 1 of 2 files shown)


src/components/About/Blog/index.js
```diff
@@ -229,7 +229,7 @@ class Blog extends Component {
   getPosts = async offset => {
     this.setState({ postRendered: false });
     let payload = null;
-    if (this.state.nextPosts.length == 0 || offset <= this.state.startPage) {
+    if (this.state.nextPosts.length === 0 || offset <= this.state.startPage) {
```
anishagg17: a pr has already been made (#3258) to fix it

akshatnitd:
@manan2501 Please claim issue before working on it. Have to close this PR.

Issue #3267, opened 2020-01-12
Password gets auto cleared if it doesn't match


anishagg17:
**Expected Behaviour**

Password should not be cleared if it doesn't match

**Steps to reproduce it**

Add wrong password while logging in

**Screenshots of the issue**

![Hnet com-image (5)](https://user-images.githubusercontent.com/43617894/72217450-c44a9000-3553-11ea-8c2e-fbdd826e8d69.gif)

**Would you like to work on the issue?**

Yes

Sulekhiya:
Are you working on the issue?

anishagg17:
If you want to work on it You can

Sulekhiya:
Okay, I am working on it!

akshatnitd:
Invalid issue. The correct behavior is clearing it.

Sulekhiya:
Okay Cool !!

PR #3283, opened 2020-01-14
Password not removed in case of wrong password.

Sulekhiya:
Fixes #[Password should not be cleared if it doesn't match #3267] 

Changes: Password is not removed in case of wrong password.

Screenshots of the change: 

Screenshot (47)


akshatnitd:
The issue was invalid and clearing the password is right behavior.

PR #3327, opened 2020-01-19
Added circular progress to various modals

### LINKED ISSUE #3319: Use Circular Progress wherever required
**Expected Behaviour**

Use `Circular Progress` wherever needed.

Yes

Dilshaad21:
Fixes #3319 

Changes: 
- Used `Circular Progress` in Modals while loading.
- Set a common `Button` ui to all modals.

Demo Link: https://pr-3327-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/components/shared/Button.js
+++ b/src/components/shared/Button.js
@@ -1,8 +1,35 @@
-import styled from 'styled-components';
+import React from 'react';
 import _Button from '@material-ui/core/Button';
+import styled from 'styled-components';
+import CircularProgress from '@material-ui/core/CircularProgress';
+import Translate from '../Translate/Translate.react';
+import PropTypes from 'prop-types';
 
-const Button = styled(_Button)`
+const StyledButton = styled(_Button)`
   width: 10rem;
 `;
+
+const Button = props => {
+  const { buttonText, isLoading, handleClick, ...otherProps } = props;
+  return (
+    <StyledButton onClick={handleClick} {...otherProps}>
+      {isLoading ? (
+        <CircularProgress size={24} color="inherit" />
+      ) : (
+        <Translate text={buttonText} />
+      )}
+    </StyledButton>
+  );
+};
+
+Button.propTypes = {
+  buttonText: PropTypes.string,
+  handleClick: PropTypes.func,
+  isLoading: PropTypes.bool,
+};
+
+Button.defaultProps = {
+  isLoading: false,
+};
 
 export default Button;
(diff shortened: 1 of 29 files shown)


Dilshaad21:
@akshatnitd Made `LoadingButton` component and replaced it with previous Button

src/components/shared/LoadingButton.js
```diff
@@ -0,0 +1,56 @@
...
+        loading: true,
+      },
+      () => {
+        this.props.onConfirm();
+      },
+    );
+  };
+
+  render() {
+    const { loading } = this.state;
```
akshatnitd: No need to keep loading in the state. Assume it to be coming in props. And also let the name be `Button`.

Dilshaad21: Should I incorporate simple Button and LoadingButton in one file itself

akshatnitd: Assume, the default value of loading to be false. Then single component would work.

Dilshaad21:
@akshatnitd I have made changes in all files where shared Button was used. And kept normal and loading buttons in one single component. Also, tell me if some changes need to be made in the naming of props of `Button` or anywhere else.

src/components/shared/Button.js
```diff
@@ -1,8 +1,60 @@
...
+  const {
+    buttonText,
+    color,
+    variant,
+    isLoading,
+    isDisabled,
+    handleClick,
+    label,
+    style,
+  } = props;
```
akshatnitd: Use `const {loading, ...otherProps} = props;`

src/components/shared/Button.js
```diff
@@ -1,8 +1,60 @@
...
+  return (
+    <StyledButton
+      variant={variant}
+      color={color}
+      onClick={handleClick}
+      disabled={isLoading || isDisabled}
+      label={label}
+      style={style}
```
akshatnitd: Pass `{...otherProps}` to the component here. This way we can ensure all props provided by MUI can be passed down to the button

Dilshaad21:
@akshatnitd made changes, passed mui props using `otherProps`.

src/components/shared/Button.js
```diff
@@ -1,8 +1,47 @@
...
+    isLoading,
+    isDisabled,
+    handleClick,
+    ...otherProps
+  } = props;
+  return (
+    <StyledButton
+      onClick={handleClick}
+      disabled={isLoading || isDisabled}
+      {...otherProps}
```
akshatnitd: `disabled` is already a Mui prop. You should use that instead of `isDisabled`.

Dilshaad21:
@akshatnitd changes are done!

PR #3357, opened 2020-01-22
Made bot wizard stepper linear

### LINKED ISSUE #3356: Improve create bot page
**Actual Behaviour**

While creating a bot user can pass to next steps without filling the current one.

**Expected Behaviour**

User should not be allowed to pass to next steps without filling the current one.

**Screenshots of the issue**

Screenshot from 2020-01-21 23-54-01

Yes

Dilshaad21:
Fixes #3356 

Changes:
- Made the stepper in botwizard linear to avoid steps being skipped and keep a track of the steps completed.
- Functions `check` and `setStep` are no longer required as user would be passing through the steps in order.

Demo Link: https://pr-3357-fossasia-susi-web-chat.surge.sh

Screenshots of the change: 

Screenshot from 2020-01-22 11-36-10


Dilshaad21:
When `Stepbutton` were allowed user could skip the steps leaving fields like `botName` or without passing some steps.

tobehonest:
Looks good now :+1:

Dilshaad21:
@akshatnitd PTAL

src/components/cms/BotBuilder/BotWizard.js
```diff
@@ -540,20 +530,6 @@ class BotWizard extends React.Component {
     }
   };
 
-  check = () => {
```
akshatnitd: How is this check handled now?

Dilshaad21: The check will no longer be required, earlier it was used to check whether the user has passed the `Configure` tab before deploying. Now user will be allowed to go to `Deploy` through it, anyways I have added the check in `handleNext`

PR #3358, opened 2020-01-22
resolved avatar rendering issue

### LINKED ISSUE #3350: Bug in passed avatar link for image preview
**Actual Behaviour**

Currently, the avatar link passed to get `avatarThumbnail` is having bugs. As a result, the image is not rendered properly

**Expected Behaviour**

Resolve the avatar link to get the correct image preview url wherever its used.

**Screenshots of the issue**

Screenshot from 2020-01-21 15-25-24

Yes

Dilshaad21:
Fixes #3350 

Changes: 
- Initially, the avatar was uploaded in `base64` image format due to which the avatar was not rendered properly at some places.
- Now using `urlToFile` function the preview url is directly converted to file which solves the problem.
- Changes were made at two places `skillWizard` and `accountTab` for user avatar and bot image uploads.
- Changes would be visible in user avatars of skill Feedback section as well.
 
Demo Link: https://pr-3358-fossasia-susi-web-chat.surge.sh

Screenshots of the change: 
**Before**
Screenshot from 2020-01-22 13-39-32

**After**
Screenshot from 2020-01-22 13-38-33


Dilshaad21:
@akshatnitd PTAL

Dilshaad21:
I think it's happening because the state `isAvatarAdded` and `isAvatarUploaded` is not getting updated sometimes. Now, I used a callback to execute an action only after the state has got updated

akshatnitd:
@Dilshaad21 The issue still persists.

Dilshaad21:
@akshatnitd can you explain when its happening exactly

Dilshaad21:
@akshatnitd resolved now! Actually the issue was with `settingsSave` which need to be false once an avatar is uploaded. Thanks.

PR #3377, opened 2020-01-23
clean up subscriptions

### LINKED ISSUE #3376: Subscriptions cleanup required
**Expected Behaviour**

Subscriptions cleanup required is required to prevent memory leaks

**Screenshots of the issue**
Screenshot 2020-01-23 at 11 21 33 PM

Yes

anishagg17:
Fixes #3376 

Changes: 

cancelled all subscriptions and asynchronous tasks in the componentWillUnmount method.

Reference Link : 

 https://github.com/material-components/material-components-web-react/issues/434#issuecomment-449561024

Demo Link: https://pr-3377-fossasia-susi-web-chat.surge.sh


anishagg17:
WIP

anishagg17:
I have fixed it in all places where it was required

anishagg17:
I have fixed it in all places where it was required

anishagg17:
@akshatnitd please review this pr

anishagg17:
> This problem won't be there once we move to hooks

But sir for that state management would be tough

anishagg17:
State is updated after unmounting as cDM fetches data and executes state update.
 as we switch pages the process of setting state continues in background hence this error arrises

anishagg17:
I have checked it on various sites and this is the only solution they all come up with

anishagg17:
@akshatnitd I have found another way to fix this issue ,i.e, By disabling state updates for each component
after they are unmounted,like

 `
componentWillUnmount() {
    this.setState = () => {};
  }
`
should I make use of this?

src/components/cms/Dashboard/MyAnalytics.js
```diff
@@ -26,24 +28,35 @@ class MyAnalytics extends Component {
...
     try {
       // eslint-disable-next-line camelcase
       let payload = await fetchSkillsByAuthor({ author_email: email });
       this.saveUsageData(payload.authorSkills || []);
-      this.setState({
-        loading: false,
-      });
+      if (this._isMounted) {
+        this.setState({
+          loading: false,
```
anishagg17: @akshatnitd the warning is given by this too.. I don't think it's a good practice to bring loading state too into redux store or parent

anishagg17: You may visit #3473 for more

PR #3440, opened 2020-02-08
fixed error in chat with susi

### LINKED ISSUE #3441: site breaks on chating with susi
**Actual Behaviour**

site breaks on chating with susi

**Steps to reproduce it**

try chatting with susi

**Screenshots of the issue**

Hnet com-image

Yes

anishagg17:
Fixes #3441

Changes: fixed error in chat with susi 

Demo Link: https://pr-3440-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/components/ChatApp/MessageListItem/Feedback.react.js
+++ b/src/components/ChatApp/MessageListItem/Feedback.react.js
@@ -27,10 +27,13 @@ class Feedback extends React.Component {
     theme: PropTypes.string,
   };
 
-  state = {
-    feedbackInProgress: false,
-    skill: this.parseSkill(),
-  };
+  constructor(props) {
+    super(props);
+    this.state = {
+      feedbackInProgress: false,
+      skill: this.parseSkill(),
+    };
+  }
 
   parseSkill = () => {
     const { message } = this.props;


src/components/ChatApp/MessageListItem/Feedback.react.js
```diff
@@ -50,6 +45,11 @@ class Feedback extends React.Component {
     return skillObj;
   };
 
+  state = {
+    feedbackInProgress: false,
+    skill: this.parseSkill(),
```
anishagg17: earlier parseSkill() was called before being declared

Dilshaad21: It will work but I don't think it's a good way of declaring state after any function as its a basic structure for components to declare state above all. We can look for some other option in this case.

Dilshaad21:
Setting the `skill` in `componentDidMount` will be an option.

anishagg17:
> Setting the `skill` in `componentDidMount` will be an option.

let's wait for mentor to suggest what to do next

src/components/ChatApp/MessageListItem/Feedback.react.js
```diff
@@ -50,6 +45,11 @@ class Feedback extends React.Component {
     return skillObj;
   };
 
+  state = {
```
akshatnitd: Please put state declaration in this case inside the constructor.

anishagg17: okay

PR #3454, opened 2020-02-11
turned many elements into styled components

### LINKED ISSUE #3453: turn into styled components
**Expected Behaviour**

there are various components which need to be turned into styled components
 

 

YEs

anishagg17:
Fixes #3453

Changes:
turned many elements into styled components

Demo Link: https://pr-3454-fossasia-susi-web-chat.surge.sh


src/components/About/Overview/index.js
```diff
@@ -41,6 +41,10 @@ const Button = styled(_Button)`
   margin: 4px;
 `;
 
+const Link = styled(_Link)`
```
akshatnitd: Put styled-component which is used multiple times into a single file.

anishagg17: @akshatnitd changes have been done

PR #3523, opened 2020-05-06
Fixed: blog page doesn't renders on accessing through team page 

### LINKED ISSUE #3521: Blog page doesn't renders on accessing through team page
**Actual Behaviour**

When user go on the team page and afterwards tap on the blog button from the footer segment, then it does not renders the page, but on reloading it does.

**Expected Behaviour**
On clicking blog button, it will render the page

**Steps to reproduce it**
Go to the team page.
Now, click on blog button in footer section.

Yes

1209simran:
Fixes #3521 

Changes: changed Link tag with anchor tag

Demo Link: https://pr-3522-fossasia-susi-web-chat.surge.sh/

### DIFF
--- a/src/components/Footer/Footer.react.js
+++ b/src/components/Footer/Footer.react.js
@@ -124,7 +124,7 @@ const Footer = () => {
                 <Link to="/devices">Devices</Link>
               </li>
               <li>
-                <Link to="/blog">Blog</Link>
+                <a href="/blog">Blog</a>
               </li>
               <li>
                 <a href={urls.API_URL}>API</a>


src/components/Footer/Footer.react.js
```diff
@@ -124,7 +124,7 @@ const Footer = () => {
                 <Link to="/devices">Devices</Link>
```
maze-runnar, line 124: every others are rendering with Link to why <a> for this one! any specific reasons for this change

1209simran: This page need to be reloaded as Link tag doesn't do so, it doesn't update the state.
So, either we need to use window.location.reload() or simply can use anchor tag.

PR #3549, opened 2020-05-14
Fixed:Contributors section is empty in the Team Page when a user is logged in

### LINKED ISSUE #3547: Contributors section is empty in the Team Page when a user is logged in
**Actual Behaviour**

When the user is logged in and visits Team page then the Contributors section is empty but it shows up when the user is not logged in.

**Expected Behaviour**

The Contributors section must show the list of contributors both when the user is logged in as well as logged out.

**Steps to reproduce it**

Login to Susi.ai
Go to the team page.
Now, visit the Contributors section at bottom of the team page.

**LogCat for the issue**

image

**Screenshots of the issue**

image

Yes, I would like to work on the issue.

1209simran:
Fixes #3547 

Changes:
Used Axios in place of ajax. On using ajax, access token is added in the url when the user is logged in due to which contributors section gets empty.

Demo Link: https://pr-3547-fossasia-susi-web-chat.surge.sh

Screenshots of the change:
DeepinScreenshot_20200514180703

### DIFF
--- a/src/apis/index.js
+++ b/src/apis/index.js
@@ -2,6 +2,7 @@
 import ajax from '../helpers/ajax';
 import urls from '../utils/urls';
 import _get from 'lodash/get';
+import axios from 'axios';
 
 const { API_URL, SOUND_SERVER_API_URL, GITHUB_API } = urls;
 const AUTH_API_PREFIX = 'aaa';
@@ -102,6 +103,7 @@ export function fetchDevices(payload) {
 
 export async function getContributors() {
   let data = [];
+  let contributors = [];
   let repos = [
     'susi.ai',
     'susi_server',
@@ -113,11 +115,11 @@ export async function getContributors() {
   try {
     let res = await Promise.all(
       repos.map(repo =>
-        ajax.get(`https://api.github.com/repos/fossasia/${repo}/contributors`),
+        axios.get(`https://api.github.com/repos/fossasia/${repo}/contributors`),
       ),
     );
-
-    res.forEach(value => {
+    res.map((response, index) => contributors.push(response.data));
+    contributors.forEach(value => {
       for (let key in value) {
         if (key.toString() !== 'statusCode') {
           let contributor = value[key];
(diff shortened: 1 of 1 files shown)


src/apis/index.js
```diff
@@ -2,6 +2,7 @@
 import ajax from '../helpers/ajax';
```
arteevraina, line 2: Since, `ajax` is not being used. Can we remove it from the imports?

927tanmay: > Since, `ajax` is not being used. Can we remove it from the imports?
No, @arteevraina we can't remove import ajax statement since other functions in the code are using it. I think the pr genrated is absolutly fine.

arteevraina: @927tanmay Yes, I was wrong.

src/apis/index.js
```diff
@@ -113,11 +115,11 @@ export async function getContributors() {
   try {
     let res = await Promise.all(
       repos.map(repo =>
-        ajax.get(`https://api.github.com/repos/fossasia/${repo}/contributors`),
+        axios.get(`https://api.github.com/repos/fossasia/${repo}/contributors`),
```
anishagg17, line 118: try using `ajax` instead of `axios`. As `ajax` function is based on `axios` with some modified headers, it is not the standard ajax that is directly imported form some library

1209simran: In `ajax` file, it is adding the access token to the url for get method, due to which url changes for the api of contributors and returns an empty array.
Could you please elaborate, what's the problem in using axios here?

anishagg17: try using 
```
repos.map(repo => {
  const settings={ isTokenRequired: false };
  return axios.get(url,null,settings)
})
```

you can see more settings at 

https://github.com/fossasia/susi.ai/blob/5c698dba1258d6eb17296cfb3f39451ef9783033/src/helpers/ajax.js#L13

1209simran: We can do one more thing, we can add if condition for this url so that token would not be add in this url.

1209simran: Which method should I prefer?

1209simran:
@akshanitd , kindly merge the PR.

1209simran:
@akshatnitd , kindly merge the PR.

Issue #3550, opened 2020-05-14
Skilll checkbox getting checked automatically


arteevraina:
**Actual Behaviour**

When more than three `checkboxes` are `unchecked` skill `checkbox` gets automatically `checked`.

**Expected Behaviour**

It should be restricted.

**Screenshots of the issue**

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/43968121/81982276-2f65b980-964f-11ea-9af9-b249a848d884.gif)

**Would you like to work on the issue?**

Yes!

arteevraina:
@snitin315 @akshatnitd What should be the expected behavior?

927tanmay:
@arteevraina I think its not an issue rather its an feature that is being implemented in both the skill checkbox and the language checkbox. Since you need to have atleast one of the four things out of Skill, Description, Examples or Author selected for a relevant search.

arteevraina:
@927tanmay So, I think these lines of code should not be there then ?
```
 if (selected.length === 0) {
      return <DashIcon style={{ marginTop: '4px', color: '#565656' }} />;
    }
```

snitin315:
This is expected behavior Skill is default

BLasan:
@mariobehling Is this still opened?

PR #3551, opened 2020-05-15
refactor: removed if condition for no checkbox selected

arteevraina:
Closes: #3550 
Changes: Removed if condition when no checkbox is selected.
@snitin315 @927tanmay Since, discussed with about Skill being default. Therefore, we don't require this if condition when nothing is checked.
Please Review :)

### DIFF
--- a/src/components/NavigationBar/SearchBar.js
+++ b/src/components/NavigationBar/SearchBar.js
@@ -6,7 +6,6 @@ import OutlinedInput from '@material-ui/core/OutlinedInput';
 import MenuItem from '@material-ui/core/MenuItem';
 import Checkbox from '@material-ui/core/Checkbox';
 import ListItemText from '@material-ui/core/ListItemText';
-import DashIcon from '@material-ui/icons/Remove';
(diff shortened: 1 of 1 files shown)


arteevraina:
> It is the desired behaviour. So, we don't need to remove this. @akshatnitd What do you think about it?

@anishagg17 This piece of code is not required. Since we always want Skill selected by default.

927tanmay:
> Closes: #3550
> Changes: Removed if condition when no checkbox is selected.
> @snitin315 @927tanmay Since, discussed with about Skill being default. Therefore, we don't require this if condition when nothing is checked.
> Please Review :)

We discussed that the issue wasn't a valid one since skill being default is the desired behaviour. So, I don't thing the mentioned line of code should be removed.

arteevraina:
> Since there should always be one option that is checked, the issue described is not valid. And you have already been told that.
> 
> But we don't want "Skills" checkbox to be checked ALWAYS. So, what are you really trying to do?

Since one operation is always checked therefore we are not going into that condition anyways that I have removed.@MuhammadWasif

PR #3553, opened 2020-05-22
Adding space in between the skill bubble on the skill page in mobile view

### LINKED ISSUE #3552: Adding space in between the skill bubble on the skill page in mobile view
**Actual Behaviour**

The bubbles are sticking above each other on the skill page in mobile view and also the width of the bubble section is abruptly fixed.

**Expected Behaviour**

There should be space in between them and also the width of that section needs to be perfectly fixed.

**Steps to reproduce it**

1. Click on any skill card.
2. See the error.

**Screenshots of the issue**

WhatsApp Image 2020-05-22 at 22 17 45

Yes, I'm working on it.

1209simran:
Fixes #3552 

Changes: Added space in between the skill bubbles and width of the skill bubble section changed to 100%.

Demo Link: https://pr-3552-fossasia-susi-web-chat.surge.sh

Screenshots:
**Before**
WhatsApp Image 2020-05-22 at 22 17 45

**After**
WhatsApp Image 2020-05-22 at 23 29 10

### DIFF
--- a/src/components/cms/SkillPage/SkillListing.js
+++ b/src/components/cms/SkillPage/SkillListing.js
@@ -55,7 +55,7 @@ const Paper = styled(_Paper)`
 `;
 
 const ExampleWrapper = styled.div`
-  width: 55%;
+  width: 100%;
   display: flex;
   flex-wrap: wrap;
   flex-direction: row;
@@ -359,7 +359,7 @@ class SkillListing extends Component {
                       key={index}
                       data={data}
                       history={history}
-                      margin={'1.5% 2.85% 1.5% 0'}
+                      margin={'12px 10px 12px 0'}
                     />
                   );
                 })}


1209simran:
@akshatnitd , i've made the changes in desktop view too. Changed the width from 55% to 100%.

src/components/cms/SkillPage/SkillListing.js
```diff
@@ -359,7 +359,7 @@ class SkillListing extends Component {
                       key={index}
                       data={data}
                       history={history}
-                      margin={'1.5% 2.85% 1.5% 0'}
+                      margin={'12px 10px 12px 0'}
```
anishagg17, line 362: Why `pixles` insted of `percentage` ?

1209simran: Percentage is a relative unit. So, the margin will change according to the screen of size.On a really small screen like a phone that might be appropriate. But if you're viewing that same page on a bigger size screen, that margin is going to be huge if we use percentage. That's why I've used pixels to maintain the same gap between them.

1209simran:
@akshatnitd , please suggest if there are some more changes.

PR #3558, opened 2020-05-23
Fixed: About,Devices,Support,Blog page doesn't renders from footer through Team Page 

### LINKED ISSUE #3557: About,Devices,Support,Blog page doesn't renders from footer through Team Page
**Actual Behaviour**

When a user goes on the team page and afterwards tap on the About, Devices, Blog or Support button from the footer segment, then it does not renders the respective pages, but on reloading it does.

**Expected Behaviour**

The About, Devices, Blog and Support page must be rendered on clicking their respective button from the footer segment on accessing through the Team page 

**Steps to reproduce it**

Go to the Team page.
Now, click on the About, Devices, Blog or Support button from the footer section.

Yes, I would like to work on the issue.

927tanmay:
Fixes #3557  

Changes: Changed the Link tags in the footer with the hyperlink tags.

Demo Link: https://pr-3557-fossasia-susi-web-chat.surge.sh

### DIFF
--- a/src/components/About/Team/index.js
+++ b/src/components/About/Team/index.js
@@ -138,7 +138,7 @@ const Team = props => {
     document.title =
       'Developer Team of SUSI.AI - Open Source Artificial Intelligence for Personal Assistants, Robots, Help Desks and Chatbots';
 
-    async function fetchData() {
+    (async function fetchData() {
       const response = await getContributors();
       let flags = [],
         output = [];
@@ -150,8 +150,7 @@ const Team = props => {
         output.push(response[i]);
       }
       updataContributors(output);
-    }
-    fetchData();
+    })();
   }, []);
 
   const createContributorCard = (contributor, key) => {
--- a/src/components/About/Team/index.js
+++ b/src/components/About/Team/index.js
@@ -133,26 +133,26 @@ const Description = styled.div`
 
 const Team = props => {
   let [contributors, updataContributors] = useState([]);
-  /*eslint-disable */
 
-  useEffect(async () => {
-    // Adding title tag to page
+  useEffect(() => {
     document.title =
       'Developer Team of SUSI.AI - Open Source Artificial Intelligence for Personal Assistants, Robots, Help Desks and Chatbots';
 
-    let data = await getContributors();
-    let flags = [],
-      output = [];
-    for (let i = 0; i < data.length; i++) {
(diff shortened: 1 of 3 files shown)


src/components/About/Team/index.js
```diff
@@ -133,26 +133,26 @@ const Description = styled.div`
...
     document.title =
       'Developer Team of SUSI.AI - Open Source Artificial Intelligence for Personal Assistants, Robots, Help Desks and Chatbots';
 
-    let data = await getContributors();
-    let flags = [],
-      output = [];
-    for (let i = 0; i < data.length; i++) {
-      if (flags[data[i].name]) {
-        continue;
+    async function fetchData() {
```
anishagg17, line 141: here `IIFE` can be used

927tanmay: Using this also have the same effect as of IIFE, for using async function in hooks this is the recommended way. We can use either of them. Now it solves the problem of without having the need to reload the page.

surge_deploy.sh
```diff
@@ -17,3 +17,12 @@ export SURGE_TOKEN=a34fd7119aacd204150edd6a8e29903d
 
 export DEPLOY_DOMAIN=https://pr-${TRAVIS_PULL_REQUEST}-fossasia-susi-web-chat.surge.sh
 surge --project ./build --domain $DEPLOY_DOMAIN;
+
```
akshatnitd, line 20: Remove these extra blank lines. Rest looks good!

927tanmay: @akshatnitd please check again I have removed the extra spaces

src/components/About/Team/index.js
```diff
@@ -133,26 +133,25 @@ const Description = styled.div`
...
     document.title =
       'Developer Team of SUSI.AI - Open Source Artificial Intelligence for Personal Assistants, Robots, Help Desks and Chatbots';
 
-    let data = await getContributors();
-    let flags = [],
-      output = [];
-    for (let i = 0; i < data.length; i++) {
-      if (flags[data[i].name]) {
-        continue;
+    (async function fetchData() {
```
anishagg17, line 141: Why to name the function, `fetchData` is a very common function name and can be used for future.

927tanmay: @anishagg17 please review I have changed the function the name.

PR #3566, opened 2020-06-02
Icons Added in language menu

### LINKED ISSUE #3565: New UI for the Language Select Menu which would include 'add' and 'close' icons
**Actual Behaviour**

Currently when a user clicks on a language option it just highlights itself. It will be great if we can provide the menu with dynamic icons.

**Expected Behaviour**

We can add some icons which changes accordingly. A 'x' icon will be displayed in front of a selected language option and similarly a '+' in front of unselected options which would change dynamically upon toggling the selections

Screenshot (123)

Yes, I would love to work on the issue

Akhil-Kashyap:
Fixes #3565 

Changes: Respective Icons were added to the Menu which changes accordingly.

**Before**
Screenshot (123)

**After**
after_pull


src/components/NavigationBar/index.js
```diff
@@ -103,7 +105,7 @@ const SusiLogo = styled.img`
     props.marginRight &&
     css`
       margin-right: ${props => props.marginRight + 'px'};
-    `}
+    `};
```
anishagg17, line 108: prettier issue

Akhil-Kashyap:
> Better UX would be similar to the one shown in the 2nd example in this link - https://material-ui.com/components/selects/#multiple-select, which uses a checkbox.



> Better UX would be similar to the one shown in the 2nd example in this link - https://material-ui.com/components/selects/#multiple-select, which uses a checkbox.

Sure, I'll look forward to implement this UI.

Akhil-Kashyap:
Screenshot (126)

I have implemented the following UI . Please have a look @akshatnitd

Akhil-Kashyap:
Done with the changes.
Screenshot (127)

PR #3592, opened 2020-08-16
footer on support page made responsive

### LINKED ISSUE #3561: Un-responsive footer on support page
**Actual Behaviour**
On the support page the footer is not responding i.e. all links are unclickable.

**Expected Behaviour**
The footer should be working normally as it does on other pages.

**Steps to reproduce it**
- Go to the support page by selecting it from footer.
- Try clicking any footer links on this page.

**Screenshots of the issue**

support-footer-error

Yes!

rajprakash00:
Fixes #3561 

Changes: Fixed the footer on support page which had extra padding due to which route to other links weren't working.

Demo Link: https://fossasia-footer.netlify.app/

### DIFF
--- a/src/components/About/Support/index.js
+++ b/src/components/About/Support/index.js
@@ -185,8 +185,7 @@ const SignUpFooter = styled.div`
 
   @media (min-width: 1400px) {
     position: relative;
-    top: 6.25rem;
-    padding: 0 0 9.375rem 0;
+    padding: 3.125rem 0 3.125rem 0;
   }
 
   @media (max-width: 1000px) {


src/components/About/Support/index.js
```diff
@@ -185,8 +185,7 @@ const SignUpFooter = styled.div`
 
   @media (min-width: 1400px) {
     position: relative;
-    top: 6.25rem;
-    padding: 0 0 9.375rem 0;
+    padding: 3.125rem 0 3.125rem 0;
```
elit-altum, line 188: This is a hard coded solution and doesn't work on various smaller screen sizes. I think you can either add more media queries tracking those breakpoints or find another way around.

elit-altum (changes requested):

src/components/About/Support/index.js
rajprakash00: I don't think it's hard-coded it was actually already prewritten [here](https://github.com/fossasia/susi.ai/blob/68f770103ca50c4acfe5b3a728218194f2886e78/src/components/About/Support/index.js#L184)(for all sizes) & I just rectified it in min width(for sizes greater than 1400px) since the padding-down & top value was mistakenly more so was overlapping with links,so basically no problem for smaller sizes.(i have checked it too)

Dilshaad21:
@rajprakash00 please share the screenshots of the changes you have made.

rajprakash00:
@Dilshaad21 I've attached the screenshots 
Actually there's no major change to the UI(so i guess screenshot would be of no use) ,that's why I have given the demo link in my PR

Dilshaad21 (approved):

PR #3617, opened 2020-12-18
Recaptcha issue fixed (#3564)

### LINKED ISSUE #3564: Issue in recaptcha while signing up
**Actual Behaviour**

Recaptcha doesn't load and shows error on the actual site

**Expected Behaviour**
Recaptcha should load and I should be able to signup

**Steps to reproduce it**
>Click on Login
>Click on SIgn up for susi
>Fill in details and click the captcha checkbox

**LogCat for the issue**

**Screenshots of the issue**

Screenshot 2020-05-31 at 3 32 46 PM

Yes

ashutoshc8101:
Fixes #3564 

Changes: 
- Fixed recaptcha not working issue at sign up page

What was wrong:
 - Instead of linking google recaptcha library with script tag, code inside library was copied and stored locally in recaptcha.js
 - But google updated their recaptcha library with time but our copy of recaptcha library didn't got updated with theirs.
 - If instead of copying their library, we would have linked it with script tag, then our application would have got access to latest version of google's library.  
 - The version of library which was copied to project got depreciated and broke the signup code.
 - Google even writes `PLEASE DONOT COPY AND PASTE THIS CODE` in the begining of their recaptcha library.
Screenshot from 2020-12-18 18-14-14

What I did:
- I removed local outdated recaptcha.js library and linked the library which google maintains with script tag in index.html
- I removed recaptcha.js from webpack entry file list

Demo Link: https://pr-3617-fossasia-susi-web-chat.surge.sh

Screenshots of the change: 
BEFORE:
Screenshot from 2020-12-18 18-14-44

NOW:

Screenshot from 2020-12-18 18-15-44


ashutoshc8101:
@akshatnitd Please review

Dilshaad21 (approved):
LGTM

akshatnitd (changes requested):
We want to keep the static js in the public folder and not load from CDN. So, please revert the change or maybe tupdate the `recaptch.js` file and see.

ashutoshc8101:
@akshatnitd Updated recaptcha.js.
Screenshot after changes:
Screenshot from 2020-12-24 20-29-26

akshatnitd (approved):

iamareebjamal:
Great way to regress the issue after a few months again, it will break again, issue will remain opened for months or years and then fixed again

ashutoshc8101:
@akshatnitd Travis CI should be setup to update recaptcta.js  everytime whenever project is deployed.
So we can have updated version of library without having to use CDNs.

akshatnitd:
@iamareebjamal I agree it is a short-term fix, but it will now unblock users from signing up.

@ashutoshc8101 Yes, we can go that route and update the `recaptch.js` in the Travis CI pre-deploy step. This will help us not to load the script through Google CDN.

ashutoshc8101:
@akshatnitd 
Screenshot from 2020-12-27 16-12-32

When @mariobehling asked to substitute external dependencies , he said he had following aims:
    1. Web Client should be ready to use in local network.
    2. User should not share their IP Address and other data.
    
Goals achieved by using local library of recaptcha.js
  1. Even after copying recaptcha.js, signup doesnot work without internet.
  Screenshot is taken using webclient with local susi_server without internet
Screenshot from 2020-12-27 16-19-56
2. User still have to share their IP Address and other data with google
Screenshot with present deployment of https://susi.ai/ with internet connection
Screenshot from 2020-12-27 16-12-54

#### Conclusion:
Copying recaptcha.js doesnot solve any of the issues susi webclient had with using cdns. 
Instead this approach creates one more issue of getting outdated.

#### What should be done?
1. Own tool resembling Recaptcha library should be build and implemented in susi server and susi webclient
OR
2. Google CDN for Recaptcha.js library should be used.
OR
3. Recaptcha should be removed.

iamareebjamal:
Exactly, thank you. None of the privacy issues or internet issues are solved by copying local CAPTCHA js
That's creating new problems to solve which did not exist in the first place with extra steps.
Thinking that somehow copying the recaptcha js locally protects user privacy is insinuating that it doesn't contact Google service which is absolutely disingenuous and misleading.

The solution which does not share the data with any 3rd party and works locally is to create our own captcha which only validates a user locally or using a server we use, which is infeasible

A feasible option for not sharing the data with Google exclusively but do it with other more privacy respecting captcha service like hcaptcha which doesn't have a database of users but still, like any 3rd party will get access to the IP address of the user. 

Hence, copying the recaptcha js locally solves nothing on privacy or local dev experience front and only complicates things and breaks the site ever so often. 

Secondly, only frontend captcha without server check could keep the bots out as the API is anyway built for machines and restricting it using captcha is extremely poor decision. Now, an API user needs either a browser or app integration of Google RECAPTCHA to access the logged in API. Hence we have instead of protecting the privacy of our users, forced other clients to compromise the security of their users by forcing them to inject Google captcha in their site/app. And also, our app becomes ineligible for FDroid because of Recaptcha. How is it good for FOSS?

Hence, I vote to remove server side captcha checking and add frontend only hcaptcha to the site, without locally copying the JS.

Dilshaad21:
@iamareebjamal @ashutoshc8101 I think we should wait for @mariobehling response to this issue as well.

